PR #30, opened 2026-05-04
Add AttributeMap.

Mohammad-nassar10:
## What does this PR do?

Adds `AttributeMap` - a thread-safe, generic key-value storage component that serves as the foundation for the datastore framework. This enables plugins to store and share typed data with automatic isolation through the `Cloneable` interface.

## Why is this change needed?
Current plugin architecture lacks a mechanism for persistent data storage across requests. `AttributeMap` provides the core building block for:
- Cross-request data aggregation (e.g., latency metrics, request counts)
- Data sharing with type safety
- `Datastores` registry (PR #34), which uses `AttributeMap` as its per-topic storage unit.

## How was this tested?

- [X] Manual testing performed

## Checklist

## Related Issues
Related to #41.


nirrozenbaum:
@Mohammad-nassar10 I see it's still draft, but as a heads up - 
can we make smaller PRs? ideally PR should not be above 500 LoC unless it's a must. 
if the PRs cannot be broken into smaller PRs, that's usually a code smell and could be a red flag

Mohammad-nassar10:
> @Mohammad-nassar10 I see it's still draft, but as a heads up - can we make smaller PRs? ideally PR should not be above 500 LoC unless it's a must. if the PRs cannot be broken into smaller PRs, that's usually a code smell and could be a red flag

@nirrozenbaum Thanks for the note, it is still a work in progress, will be updated to match the requirements.

pkg/framework/datastore/attributemap.go
```diff
@@ -0,0 +1,160 @@
+/*
+Copyright 2025 The Kubernetes Authors.
```
vMaroon: Boilerplate of new files should credit `The llm-d Authors` - with the year 2026.

pkg/framework/datastore/attributemap.go
```diff
@@ -0,0 +1,114 @@
...
+	if key == "" {
+		return
+	}
+	if value == nil {
+		return
+	}
+	a.data.Store(key, value)
+}
+
+func (a *Attributes) Get(key string) (Cloneable, bool) {
```
nirrozenbaum, line 76: nit: do we really need here the bool return value?
it looks like either we return (value, true) or (nil, false) always. 
so a check if value == nil should be sufficient in the caller..

not a blocker.

pkg/framework/datastore/attributemap.go
```diff
@@ -0,0 +1,114 @@
...
+
+// Cloneable types support cloning of the value.
+// All values stored in AttributeMap must implement this interface
+// to ensure data isolation and prevent unintended mutations.
+type Cloneable interface {
+	Clone() Cloneable
+}
+
+// AttributeMap is used to store flexible metadata or traits
+// across different aspects of an inference server.
```
nirrozenbaum:
```suggestion
// across different aspects of a model.
```

PR #33, opened 2026-05-04
Add DataStore proposal.

Mohammad-nassar10:
## What does this PR do?

Adds a proposal of adding `DataStore` to `docs/proposals` documenting the
package's components (`Datastores`, `AttributeMap`, `Cloneable`), their public
APIs, thread-safety guarantees, data isolation model, and usage examples.

## How was this tested?

## Checklist

## Related Issues
Related to #41.


ronenkat (commented):
Can you merge `framework-datastore.md` and `README.md` as there is duplication between the two.
the additional details from the README can be added into the proposal as appendixes

docs/proposals/datastore-proposal/README.md
```diff
@@ -0,0 +1,240 @@
...
+- Support multiple independent data topics (e.g., `"model-latency"`, `"request-content-prefix"`) so that different data collectors do not interfere with each other.
+- Guarantee thread safety: concurrent reads and writes from different goroutines must not require external synchronization by callers.
+- Ensure data isolation: retrieving a value must return an independent copy, so callers cannot accidentally mutate shared state.
+- Keep the data layer general-purpose so it can serve future features beyond latency routing.
+
+### Non-Goals
+
+- Persistent storage (disk, database) — the datastore is in-memory only; data is lost on pod restart.
+- TTL or automatic expiration — entries remain until a data collector explicitly deletes them.
+- Distributed or shared storage across IPP replicas — each pod maintains its own independent datastore instance.
```
ronenkat:
```suggestion
- Distributed or shared storage across IPP replicas — each pod maintains its own independent datastore instance. To be revisited in a future proposal.
```

docs/proposals/datastore-proposal/README.md
```diff
@@ -0,0 +1,240 @@
...
+
+```go
+var Data *Datastores
+
+type Datastores struct {
+    mu      sync.RWMutex
+    keyName map[string]AttributeMap
+}
+
+func NewDatastores()
```
ronenkat: Lets make the newDatastores return a value that will be stored in the server.

docs/proposals/datastore-proposal/README.md
```diff
@@ -0,0 +1,240 @@
...
+    Clone() AttributeMap
+}
+```
+
+The concrete implementation (`Attributes`) uses `sync.Map` internally, providing safe concurrent access without requiring callers to hold a lock. `Get()` calls `Clone()` on the stored value before returning it, so mutations to the returned value never affect the datastore.
+
+### Datastores
+
+```go
+var Data *Datastores
```
ronenkat: lets remove the definition here, so the instance will be kept in the server code, and not be global.
Access should be done through the handle.

docs/proposals/datastore-proposal/README.md
```diff
@@ -0,0 +1,240 @@
...
+
+### Integration with Handle
+
+Plugins access the datastore through the `Handle` interface:
+
+```go
+store, err := handle.GetAttributeMap("inference-pool-latency")
+```
+
+This keeps the global `Data` variable hidden from plugins and makes the access point mockable in tests.
```
ronenkat: can drop the statement.

docs/proposals/datastore-proposal/README.md
```diff
@@ -0,0 +1,240 @@
...
+| Clone with data | Put 2 keys, Clone() | Clone has same keys/values, independent |
+| Clone independence | Clone map, modify clone | Original unchanged |
+| Get returns clone | Put value, Get twice | Two independent clones returned |
+| Update existing key | Put key="test" twice with different values | Second value overwrites first |
+| Multiple keys | Put 10 different keys | All keys retrievable, Keys() returns 10 |
+| Concurrent Put/Get/Delete | 100 goroutines doing Put, 100 doing Get, 100 doing Delete | No panics, no race conditions |
+
+**Test Rules:**
+- All tests use `testCloneableValue` struct implementing Cloneable
+- Mock no external components (datastore is self-contained)
```
ronenkat: what do you mean by - Mock no external components (datastore is self-contained) ?

ronenkat (commented):
Thank you!

docs/proposals/datastore-proposal/README.md
Mohammad-nassar10: it means that the datastore doesn't have any external dependencies that would need to be mocked in tests. rephrased.

PR #34, opened 2026-05-04
Datastore framework

Mohammad-nassar10:
## What does this PR do?

Adds `pkg/framework/datastore/datastore.go` — the top-level registry (`Datastores`)
that manages multiple named `AttributeMap` stores, each identified by a topic key.

> **Note:** This PR is stacked on #30. Because both PRs target `main`,
> the diff here includes the `attributemap.go` and `attributemap_test.go` changes from
> that PR. The net-new changes in this PR are `datastore.go` and `datastore_test.go`
> only. Please merge #30 first.

## How was this tested?

## Checklist

## Related Issues

Related to #41.


pkg/framework/datastore/datastore.go
```diff
@@ -0,0 +1,87 @@
...
+// ErrEmptyDatastoreKey is returned when a datastoreKey is empty.
+var ErrEmptyDatastoreKey = errors.New("datastore key cannot be empty")
+
+// Datastores manages multiple named data stores, each identified by a datastoreKey string.
+// Provides the top-level registry for all topic-specific AttributeMaps.
+//
+// All operations are thread-safe using RWMutex.
+type Datastores struct {
+	mu      sync.RWMutex
+	keyName map[string]AttributeMap
```
ronenkat:
```suggestion
	topic map[string]AttributeMap
```

pkg/framework/datastore/datastore.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// NewDatastores creates and returns a new Datastores instance.
+// Each caller should create and manage their own instance.
+func NewDatastores() *Datastores {
+	return &Datastores{
+		keyName: make(map[string]AttributeMap),
```
ronenkat:
```suggestion
		topic: make(map[string]AttributeMap),
```

pkg/framework/datastore/datastore.go
```diff
@@ -0,0 +1,87 @@
...
+// GetOrCreateStore returns an existing AttributeMap or creates a new one atomically.
+// Returns ErrEmptyDatastoreKey if datastoreKey is empty.
+func (ds *Datastores) GetOrCreateStore(datastoreKey string) (AttributeMap, error) {
+	if datastoreKey == "" {
+		return nil, ErrEmptyDatastoreKey
+	}
+
+	// Fast path: check if store exists with read lock
+	ds.mu.RLock()
+	store, ok := ds.keyName[datastoreKey]
```
ronenkat:
```suggestion
	store, ok := ds.topic[datastoreKey]
```

pkg/framework/datastore/datastore.go
```diff
@@ -0,0 +1,87 @@
...
+	defer ds.mu.Unlock()
+
+	// Double-check in case another goroutine created it
+	store, ok = ds.keyName[datastoreKey]
+	if ok {
+		return store, nil
+	}
+
+	store = NewAttributes()
+	ds.keyName[datastoreKey] = store
```
ronenkat:
```suggestion
	ds.topic[datastoreKey] = store
```

pkg/framework/datastore/datastore.go
```diff
@@ -0,0 +1,87 @@
...
+// No-op if the key doesn't exist.
+func (ds *Datastores) DeleteStore(datastoreKey string) error {
+	if datastoreKey == "" {
+		return ErrEmptyDatastoreKey
+	}
+
+	ds.mu.Lock()
+	defer ds.mu.Unlock()
+
+	delete(ds.keyName, datastoreKey)
```
ronenkat:
```suggestion
	delete(ds.topic, datastoreKey)
```

ronenkat (commented):
sync terms with the proposal:
keyName --> topic

pkg/datastore/datastore.go
```diff
@@ -0,0 +1,87 @@
...
+	ds.mu.RUnlock()
+	if ok {
+		return store, nil
+	}
+
+	// Slow path: create new store with write lock
+	ds.mu.Lock()
+	defer ds.mu.Unlock()
+
+	// Double-check in case another goroutine created it
```
shaneutt: Shouldn't it be `topic`? 🤔 

```go
store, ok = ds.topic[datastoreKey]
```

Mohammad-nassar10: Fixed, thanks!

nirrozenbaum (commented):
@Mohammad-nassar10 taking a step back here - 
can you explain what was your intention in topic? why is it needed? 
how would it be used? why do we have store of stores?  

we need to have datastore that maps from model (its unique identifier) to it's properties (e.g., attributesmap for dynamic collected data, metadata for static properties). 

we need to be specific and use descriptive names accordingly. 

how it should be modeled:
- add framework/datalayer dir - should contain struct `Model` with field `name` (unique identifier), and AttributesMap. 
- Datastore should have a sync.map, mapping between string(model unique id) -> Model instance (the instance includes the attributes that are collected).
- future PRs might add to the Model struct static info per model (can be added later on a need basis).

also, not related to this PR, but I see datastore dir was pushed under framework. this is wrong. (I approved previous PR but missed it). 
we should move datastore under package -> pkg/datastore. it's not part of the framework.

Mohammad-nassar10:
@nirrozenbaum The two-level design (`topic → AttributeMap`) is used because not all plugins/collectors key their data by model name (for example a prefix-caching plugin keys by prefix rule ...) so a flat `model → attributes` map breaks as soon as the plugin's primary entity isn't a model. (I think EPP ran into the same problem. Started with pod mapping then added two level map).

The `topic` key plays the role as a named namespace each plugin claims to own ("running-requests", "latency", ...), and the inner `AttributeMap` key is whatever makes sense for that plugin.

Regarding datastore dir, agreed, will move pkg/framework/datastore to pkg/datastore.

Mohammad-nassar10:
@nirrozenbaum @ronenkat Changed the datastore from a two-level `topic → AttributeMap` map to a `model → *Model map`, where each `Model` (in #55) holds an AttributeMap for runtime metrics and a place for static metadata (vendor, family, etc.) to be added in future.

pkg/datastore/datastore.go
```diff
@@ -0,0 +1,76 @@
...
+
+// Store is a thread-safe registry of Model entries keyed by model name.
+// The outer key is the model name; each Model holds an AttributeMap for
+// dynamic runtime metrics (e.g. "running-requests", "pool-latency") and
+// any static metadata added in future (e.g. vendor, family).
+//
+// All operations are thread-safe using RWMutex.
+type Store struct {
+	mu     sync.RWMutex
+	models map[string]*datalayer.Model
```
nirrozenbaum, line 40: this assumes #55 merged?
otherwise looks good. 

pay attention you need DCO (have verified signed commits)

Mohammad-nassar10: right, it is based and used the Model Struct from #55.

ronenkat, line 40: why not 
`models map[string]*AttributeMap`
do we need to duplicate the model name?

Mohammad-nassar10: I see that the commits are signed, DCO passed

pkg/datastore/datastore.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// NewStore creates and returns a new Store instance.
+func NewStore() *Store {
+	return &Store{models: make(map[string]*datalayer.Model)}
+}
+
+// GetOrCreate returns the Model for name, creating it atomically if it does not exist.
+func (s *Store) GetOrCreate(name string) *datalayer.Model {
```
ronenkat: we need an interface for reading/updating the datastore as datastore should be internal to IPP.

ronenkat:
```suggestion
func (s *Store) GetOrCreateModel(name string) *datalayer.Model {
```

pkg/datastore/datastore.go
nirrozenbaum: I think you need to rebase to solve lint issues

pkg/datastore/datastore.go
```diff
@@ -0,0 +1,76 @@
...
+	if m, ok = s.models[name]; ok {
+		return m
+	}
+	m = datalayer.NewModel(name)
+	s.models[name] = m
+	return m
+}
+
+// Delete removes a model by name. No-op if it does not exist.
+func (s *Store) Delete(name string) {
```
ronenkat:
```suggestion
func (s *Store) DeleteModel(name string) {
```

ronenkat (commented):
Nice.

pkg/datastore/datastore.go
nirrozenbaum: we don't need the duplicated model name in datastore. 
we might need it when scorers/picker get as argument (`[]datalayer.Model modelCandidates`).

Mohammad-nassar10: It is useful for adding static info in the Model Struct (if needed in the future)

pkg/datastore/datastore.go
nirrozenbaum: right 👍

Mohammad-nassar10: @ronenkat Can you explain please?

nirrozenbaum: explanation - need to add also: 
```go
type Datastore interface {
... your functions ...
}
```

this is useful to allow mock datastore in unit-tests, etc.

nirrozenbaum: can be added also in a follow up, not a blocker for current PR.

pkg/datastore/datastore.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if m, ok = s.models[name]; ok {
+		return m
+	}
+	m = datalayer.NewModel(name)
+	s.models[name] = m
+	return m
```
nirrozenbaum: lock and release in the same function is risky. plus the functionality is duplicated. 
it should be completely fine to keep it with Lock (not RLock) considering the time it takes to perform this operation is almost zero.
```suggestion
	s.mu.Lock()
	defer s.mu.Unlock()
	if m, ok = s.models[name]; ok {
		return m
	}
	m = datalayer.NewModel(name)
	s.models[name] = m
	return m
```

Mohammad-nassar10: Thanks, added the change with a small fix (`if m, ok := s.models[name]`).

pkg/datastore/datastore.go
```diff
@@ -0,0 +1,77 @@
...
+
+// store is a thread-safe registry of Model entries keyed by model name.
+// The outer key is the model name; each Model holds an AttributeMap for
+// dynamic runtime metrics (e.g. "running-requests", "pool-latency") and
+// any static metadata added in future (e.g. vendor, family).
+//
+// All operations are thread-safe using RWMutex.
+type store struct {
+	mu     sync.RWMutex
+	models map[string]*datalayer.Model
```
ronenkat, line 40: we need to decide:
either
`models []*datalayer.Model`
or 
`model map[string]*AttributeMap`
it will affect the interface.
We need to be able to change pointers so we can easily manage snapshots.
As we decided to scope down on flexibility, I would go with the first option

Mohammad-nassar10: I'd go with `map[string]*datalayer.Model` because this gives `O(1)` lookup (better than slice). And using `*Model` rather than `*AttributeMap` directly leaves room for future static fields (vendor, family, etc.) that don't come from runtime metrics.

nirrozenbaum:
testing prow commands:

/cc @shmuelk

Issue #13, opened 2026-04-27
add data preparation layer before invoking the processing plugins


nirrozenbaum:
similar to epp, it could be beneficial to have a data preparation phase to validate all registered plugins can consume the data they depend on. this could be validated using similar interfaces to epp with DataProducer/DataConsumer and DAG validation on startup.

xref: https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2618

PR #43, opened 2026-05-05
Add ModelSelector architecture proposal

noyitz:
## Summary
- Introduces the ModelSelector framework proposal for the IPP
- Adapts the upstream [Scheduler Architecture (proposal 0845)](https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/docs/proposals/0845-scheduler-architecture-proposal) Filter/Scorer/Picker pattern for model selection
- Framework-agnostic to internal/external models and K8s concepts
- Enables cost optimization, latency-based routing, and future fallback/semantic routing

## Details
The proposal defines:
- **Filter** — removes models unfit for selection (rate-limited, unavailable, auth constraints)
- **Score** — ranks remaining models with normalized [0-1] scores (cost, latency)
- **Pick** — selects winning model(s) from scored candidates
- **ModelSelector Profile** — configured set of Filters, Scorers, and a Picker
- Integration as a **RequestProcessor** in the existing IPP pipeline

Configuration API is marked as TODO and will be addressed separately.


docs/proposals/001-model-selection-framework/README.md
```diff
@@ -0,0 +1,95 @@
+# ModelSelector Architecture
```
ronenkat: I believe we should be selecting a mode and an endpoint (for inference pool).
Consider that model `gpt-oss-120b` is available on an internal server and on external server.
Hence, the same model should have two entries following the `filter` stage, and a scorer can assign different score to each one.

docs/proposals/001-model-selection-framework/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+Result: model A is injected into the request body and the request continues with the regular flow (similar flow to the one where user sends a request with a model in the body).
+
+## The connection to pluggable payload processor
+
+The current Payload Processor defines:
+- RequestProcessor
+- ResponseProcessor
+
+ModelSelector integrates as a RequestProcessor, meaning:
```
ronenkat: **Design note:**
Some scorers would need extra information than available directly in the request.
E.g., a cost scorer may need to know the complexity of the request.
It would make sense to add "extract" plugins that would be run prior to the `ModelSelector`, and the information would be passed via `CycleState`.
As a follow-up of the initial design it would be good to make sure some plugins run *before* the ModelSelector`.

ronenkat (commented):
Thank you. Great evolution.

docs/proposals/001-model-selection-framework/README.md
```diff
@@ -0,0 +1,95 @@
...
+#### Filter
+Filter runs before any scoring, and remove models that are not fit for selection. Some examples:
+- Authorization constraints
+- Rate limit exhaustion
+- Unavailable models
+
+The framework will return an error to the client if the models are filtered to zero.
+
+#### Score
+Score applies a score to each remaining model provided. Scorers SHOULD keep their score values in a normalized range: [0-1]. Any weighting should be added at the ModelSelectorProfile configuration level. Examples:
```
Mohammad-nassar10: Normalized scores require weights that are workload-dependent, hard to tune and fragile. Also, make the picker's decision opaque and not explainable.

Maybe we can think about alternative way, for example (need to be defined better and in details) consider having each scorer return a value in its natural unit (latency scorer returns milliseconds, cost scorer returns USD/1K tokens, queue scorer returns queue depth...). The picker then applies an explicit policy ("pick lowest latency", "pick lowest cost under threshold X") that is readable and explainable. One challenge is if we want to combine multiple scorers like latency with cost, this will requires the picker to do more work, but that logic is explicit rather than hidden in weight coefficients (we can do more research about case).

docs/proposals/001-model-selection-framework/README.md
nirrozenbaum: that's was not the intention here. 
think about it this way - 
model name represents a unique model across the system. the case of same model name for internal/external is not considered (they could have the same backend name, but not the same "user facing name"). 

the reasoning is simple - given a request with a specified model in the body - that should map 1:1 to a model (either external or internal). otherwise there would be ambiguity.

to make a long story short - 
model name should be unique. the model name is selected based on some scorers. if the selected model is backed by inferencepool (internal model) - the request will proceed to scheduler that will select the best endpoint to serve (given the model was already selected).

in your example - you used model-name + internal/external as the unique identifier. 
instead you can think about it as (very simplified example):
- model `gpt-oss-120b` + `internal`: rename it to `gpt-oss-120b-internal`.
- model `gpt-oss-120b` + `external`: rename it to `gpt-oss-120b-external`. 
then metrics are collected for `gpt-oss-120b-internal`/`gpt-oss-120b-external` and filter/scorer/picker produces a decision between them.

so given a request with `{"model": "auto"}` the framework would select the model. 

does it make sense?

docs/proposals/001-model-selection-framework/README.md
nirrozenbaum: pay attention that the we didn't say anything about normalized score, we say the range is normalized :). 
meaning each scorer can produce a score in the range of [0-1] (and the framework should enforce it). 

the idea behind it - if you have two scorers, each works in its own range of scores, one can easily override the other. 
example:
scorer A: produces scores in [0-1] range. 
scorer B: produces scores in [0-100] range. 

in this example, weighting of scorers becomes almost irrelevant cause scorer B will always dominate scorer A.

docs/proposals/001-model-selection-framework/README.md
nirrozenbaum: yes. I believe you're referring to this issue:
https://github.com/llm-d/llm-d-inference-payload-processor/issues/13

docs/proposals/001-model-selection-framework/README.md
```diff
@@ -0,0 +1,583 @@
...
+
+Author(s): @noyitz, @nirrozenbaum
+
+## Proposal Status
+
+***Provisional***
+
+## Summary
+
+This proposal introduces a **Model Selection Framework** for the Inference Payload Processor (IPP) that enables intelligent routing across multiple external model providers. The framework adapts the upstream [Scheduler Architecture (proposal 0845)](https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/docs/proposals/0845-scheduler-architecture-proposal) — specifically the **Filter/Scorer/Picker** pattern — for model and provider selection.
```
davidbreitgand: Is the intention to limit the scope of this proposal to external model providers only and tackle intelligent routing across both external _and_ internal models in a different proposal? Or should we change to:
```suggestion
This proposal introduces a **Model Selection Framework** for the Inference Payload Processor (IPP) that enables intelligent routing across multiple external model providers and internal models. The framework adapts the upstream [Scheduler Architecture (proposal 0845)](https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/docs/proposals/0845-scheduler-architecture-proposal) — specifically the **Filter/Scorer/Picker** pattern — for model and provider selection.
```

docs/proposals/001-model-selection-framework/README.md
davidbreitgand: > given a request with a specified model in the body - that should map 1:1 to a model (either external or internal). otherwise there would be ambiguity.

In your example, you use a convention for model naming. Is there something that can be universally accepted as a convention by the customers who might wish to continue using "standard" names? Also the same model can be offered my multiple vendors. In that case "external" is insufficient. Do you envision something like `deepseek-v3-google`, etc?

nirrozenbaum: the example was just a simplification. 
generally speaking, I'd expect the model name to uniquely represent a model across the system.
so something like:
`model name (string, the one from body) ---- represents ----> model` (could be inferencepool or external).

to allow intelligent model selection - model name must be unique. 
(the equivalent from epp is intelligent endpoint selection - endpoint must be unique, e.g, {ns, name, rank} represent an endpoint).

we have no "standard" names or naming conventions, but I do think model name uniqueness can be enforced (e.g., by validation webhooks). 

re multiple vendors of the same model - 
models should have a "user facing name" and a "backend name". we can put behind a single "user facing name" multiple external models with some traffic splitting between them. the question is if it makes sense to do so if each of them might have completely different costs/latency.

docs/proposals/001-model-selection-framework/README.md
nirrozenbaum: that was a glitch in the text. thanks for catching it!
I believe we should just phrase it as "across models". 
in upstream code, the concept of external models doesn't exist yet.
```suggestion
This proposal introduces a **Model Selection Framework** for the Inference Payload Processor (IPP) that enables intelligent routing across multiple models. The framework adapts the upstream [Scheduler Architecture (proposal 0845)](https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/docs/proposals/0845-scheduler-architecture-proposal) — specifically the **Filter/Scorer/Picker** pattern for model selection.
```

docs/proposals/001-model-selection-framework/README.md
davidbreitgand: @nirrozenbaum , @noyitz : here is another instance of accentuating **external** models.

docs/proposals/001-model-selection-framework/README.md
davidbreitgand: Sure. There are other instances in the text when you mention external models only. I'll add ditto if I spot this.

docs/proposals/001-model-selection-framework/README.md
davidbreitgand: @nirrozenbaum , I really like your thought about putting a single "user-facing" model name that maps to multiple providers! 

About the differences in cost and latency: I wonder what the actual use cases that we want to cater to are? 

In one use case, a user requesting a model from a specific provider expects unconditional routing to that provider. I think this would be evident from the URL of the request, where to route and an unconditional target model profile will not run scorers for latency and cost in the unconditional request processing pipeline. 

In another use case, the user might only be interested in the same model, but not in the provider. Then the latency and cost score can help arbitrate where to route using the user-facing model name -> provider mapping, where the scorers will only consider one mapping user-facing model --> {pool of providers} and select one provider for the user-facing model.

Finally, for the auto model UC, scorers will score across all mappings for all user-facing models and select one provider eventually. 

WDYT?

docs/proposals/001-model-selection-framework/README.md
nirrozenbaum: I think this is old commit? I don't see this instance in the PR

docs/proposals/001-model-selection-framework/README.md
nirrozenbaum: In my view, the level of abstraction here should completely detach from the terminology of external/internal and providers. 

the level of abstraction should talk in the language of `Model`, e.g.,:
```go
type Model struct {
    name string // unique name across all models
}
```
Attributes Per Model should be collected to allow scoring a model, e.g.,:
- when a request is sent (after the model is known), calculate TTFT on response hook and store it (TTFT per model). 
- when a request is sent (after the model is known), maintain in-flight active request per model.
- have a static mapping (could evolve to be dynamic if there's use case) of [$ per token] for that model name. 

then, if a request comes in for a given model - we respect it and only collect the metrics (TTFT, inflight, etc).
when a request comes in with `{"model": "auto"}` in the body, we should select the "best" model out of set of candidate models (uniquely identified, each is a unique string). each candidate model represents the user facing model name. for selecting the model we may use scorers based on the information we collected, e.g.,:
- TTFT scorer
- cost scorer 
- active requests scorer
- etc ..

what happens next is not a concern of the ModelSelector, e.g., if the client facing name represents inferencepool or if it represents external model with multiple providers.

so ModelSelector selects only the model, not provider/endpoint, nor internal/external. 
the language and abstraction level here is `model` and it's measured metrics.

ronenkat: I agree that the term model should be independent of where it is served.
The reality is that the decision factors reflect mostly the characteristic of the model serving endpoint (TTFT, number of requests, cost) -- all are determined not directly through the model concept, but rather by the serving endpoint.
It also boils down to a decision how we define a model name:
1. Is model unique? e.g., should we use `pool1/qwen` and `pool2/qwen` if `qwen` is the same model running in two locations?
2. should we allow a user to define `qwen` and IPP decides if `pool1` or `pool2` is better? 
this would be a syntactic sugar of using a group called `qwen` defined as { `pool1/qwen`, `pool2/qwen` } --- similar to the `auto` example above.
I think (2) is an extension of (1), that you describe above, and in any case IPP decisions are translated into selecting an inference pool.

usize: > Is model unique? e.g., should we use pool1/qwen and pool2/qwen if qwen is the same model running in two locations?

It complicates a number of cases where model really does need to be considered separately from endpoint. Fallback is one that comes to mind. When we fall back to a different inference pool, or from one third party provider to another, the model should be invariant -- otherwise any sort of semantic routing or complexity analysis that led to the model choice gets lost.

A clean separation between model and endpoint semantics would decrease operational overhead here. For example, if we need to rerun certain policies related to endpoint -- like credential injection -- while skipping policies related to the model alone.

This could be implemented by treating endpoint + model as a path and splitting. Though that splitting logic will have to be factored in across many policies, right?

> should we allow a user to define qwen and IPP decides if pool1 or pool2 is better?
this would be a syntactic sugar of using a group called qwen defined as { pool1/qwen, pool2/qwen } --- similar to the auto example above.
I think (2) is an extension of (1), that you describe above, and in any case IPP decisions are translated into selecting an inference pool.

This is a good point. If we have a different cost basis for a model depending on the inference provider -- definitely the case -- how could we factor that into our model picking?

Enumerating all possible combinations of `{endpoint} / {model}` is one possibility. That said, we can also simply represent the costs as a mapping of `{model} => [inference_provider0_cost, ... , inference_providerN_cost]` and consider that while keeping model separate from the provider.

docs/proposals/001-model-selection-framework/README.md
```diff
@@ -0,0 +1,95 @@
...
+- The entry & exit points are defined by the framework, forming a clear API surface
+- State management
+  - Per-request state is managed via CycleState and shared across plugins during a single selection run
+- The system should be extensible via pluggable components (Filter, Score, Pick)
+
+## Definitions
+- **ModelSelector framework** – A pluggable system for implementing model selection logic
+- **ModelSelector Profile** – A configured set of Filter(s), Scorer(s), and Picker
+- **ModelSelector Profile Run** – A single execution of Filter(s), Scorer(s), & Picker used to select models per request.
+- **ModelCandidate** – A concrete model option eligible to serve a request.
```
shaneutt: I can't find anything in this proposal where we define who creates `ModelCandidates`, where they come from.

docs/proposals/001-model-selection-framework/README.md
```diff
@@ -0,0 +1,95 @@
...
+- ResponseProcessor
+
+ModelSelector integrates as a RequestProcessor, meaning:
+- It can run alongside other request processors
+- It composes naturally within the existing pipeline
+- It can leverage CycleState to share data across plugins
+
+## Configuration API
+Configuration API should be defined for PayloadProcessor in general and should address ModelSelector as well.
+<!-- TODO -->
```
shaneutt: Are we planning to include the user facing config API as part of this PR, or some future iteration?

docs/proposals/001-model-selection-framework/review-feedback.md
```diff
@@ -0,0 +1,127 @@
...
+### ModelSelector Profile
+
+A Profile consists of 3 phases: `Filter`, `Score`, and `Pick`.
+
+*Profile Constraints*
+- A profile can have any number of `Filter` plugins registered (including zero)
+- A profile can have any number of `Score` plugins registered (including zero)
+- A profile MUST have exactly one `Pick` plugin registered
+
+Multiple profiles may be defined, allowing different selection strategies to be applied based on request characteristics.
```
shaneutt: We should spell out some specific selection strategies here.

docs/proposals/001-model-selection-framework/review-feedback.md
```diff
@@ -0,0 +1,127 @@
...
+Filter runs before any scoring, and removes models that are not fit for selection. Some examples:
+- Authorization constraints
+- Rate limit exhaustion
+- Unavailable models
+
+The framework will return an error to the client if the models are filtered to zero.
+
+#### Score
+Score applies a score to each remaining model. Scorers SHOULD keep their score values in a normalized range [0-1]. Any weighting should be added at the Profile configuration level. Scorers may use both static configuration and runtime metrics from the ModelCandidate to compute scores. Examples:
+- Cost-based scoring (using per-token cost from static config)
```
davidbreitgand: This [recent paper](https://arxiv.org/pdf/2603.23971) by Stanford, Berkeley, CMU, and Microsoft Reasearch, suggests that for reasoning models the cost is dominated by thinking tokens billed as part of the output tokens. This leads to price reversals when for the same prompt a nominally cheaper model ends up to be dramatically more costly. The architecture should allow to collect actual costs and feed them into the scorer. Probably, this is will not change the design of the scorer itself, but static configuration should not be the only way to obtain costs. 

Also, Anthropic, Google, and OpenAI allow explicit, hybrid, and implicit KV cache maintenance in the cloud. There is a 90% discount on the input tokens by these providers. The cache will be managed outside of the Filter->Scorer->Picker, but the cost is not static. Rather it depends, e.g., on whether a KV cache exists at a specific provider for this prompt prefix. So, again based on the context, the actual cost for this request might be dramatically different from a nominal static cost.

docs/proposals/001-model-selection-framework/README.md
```diff
@@ -0,0 +1,95 @@
...
+  - Opinions should be held by the plugins, not the framework
+- The entry & exit points are defined by the framework, forming a clear API surface
+- State management
+  - Per-request state is managed via CycleState and shared across plugins during a single selection run
+- The system should be extensible via pluggable components (Filter, Score, Pick)
+
+## Definitions
+- **ModelSelector framework** – A pluggable system for implementing model selection logic
+- **ModelSelector Profile** – A configured set of Filter(s), Scorer(s), and Picker
+- **ModelSelector Profile Run** – A single execution of Filter(s), Scorer(s), & Picker used to select models per request.
```
nirrozenbaum:
```suggestion
- **ModelSelector Profile Run** – A single execution of Filter(s), Scorer(s), & Picker used to select model(s) per request.
```

docs/proposals/001-model-selection-framework/README.md
nirrozenbaum: future iteration. config api is needed in general for IPP and not specific for this proposal. 
the future design and implementation of that config API should include model selector as well.

docs/proposals/001-model-selection-framework/review-feedback.md
```diff
@@ -0,0 +1,127 @@
+# Proposal Review Feedback
+
+Nir, here's my feedback on the rewrite. Overall direction is right — lean, upstream-friendly, framework-agnostic. Below are the items I'd address, followed by an updated version with all of them applied.
```
nirrozenbaum: claude generated, should be removed :).
let's iterate and sync the required changes into the proposal README.md.

docs/proposals/001-model-selection-framework/README.md
nirrozenbaum: @noyitz not related to the proposal text - please rename to:
`docs/proposals/043-model-selection-framework/README.md`.
proposal dirs are matching the PR number for tracking.

nirrozenbaum: IMO we should start simple. 
the way I interpret this thread - it is all a matter of how we choose to represent things. 
when I say model, my intention is a model that can be tracked in terms of costs, latency, accuracy, etc. 
if a model has different costs on different providers, we should treat that as different models. at least in the first iteration. otherwise we complicate things instead of starting simple.

to answer the questions here, in my view (at least for first iteration): 
- yes, model name should be considered unique.
- endpoints should be left out of scope. think about the use case where you send your inference request to `gpt5` on `api.openai.com`. there's no endpoint in the picture (although openai has a lot of endpoints behind the scenes).

docs/proposals/001-model-selection-framework/README.md
nirrozenbaum: added a short section. that would of course have to evolve and include information like where do they come from.

docs/proposals/001-model-selection-framework/review-feedback.md
nirrozenbaum: removed it. it's not part of the first iteration of the proposal (future enhancement).

docs/proposals/001-model-selection-framework/review-feedback.md
nirrozenbaum: agreed. I don't think there's any intention to select model based on static configuration, otherwise we don't really need pluggable framework :).

static cost was suggested just as a simple example to start with, and it can evolve to be dynamic attribute collected by the framework.

docs/proposals/043-model-selection-framework/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+While the upstream scheduler selects an **endpoint** for a request, the ModelSelector Framework selects a **model** to serve a request. This is based on configuration policies and enables cost optimization, latency-based routing, and future fallback and semantic routing capabilities.
+
+## Design Principles
+- The *framework* should be agnostic to model types and K8s concepts.
+  - Opinions should be held by the plugins, not the framework
+- The entry & exit points are defined by the framework, forming a clear API surface
+- The system should be extensible via pluggable ModelSelector components (Filter, Score, Pick)
+- State management
+  - Per-request state is managed via in-memory state that is shared across plugins during a single selection run
```
ronenkat, line 19:
```suggestion
  - Per-request state is managed via in-memory state that is shared across plugins during a single selection run
  - Global state is managed by the IPP data layer, and is shared with the plugin
```

docs/proposals/043-model-selection-framework/README.md
```diff
@@ -0,0 +1,95 @@
...
+- Latency-based scoring (using observed response times from runtime metrics)
+
+#### Pick
+Picker selects the model(s) from the provided list of scored models. Picker MUST return one model at minimum. Examples:
+- Highest score wins
+- Weighted random sampling (use weighted score as the probability to win)
+
+### Metrics and Runtime Feedback
+
+Filters and Scorers may consume runtime metrics collected by ResponseProcessor plugins and stored in-memory. This enables data-driven selection decisions based on observed model behavior (latency, error rates, rate limit status) rather than static configuration alone.
```
ronenkat:
```suggestion
Filters and Scorers may consume runtime metrics from the data layer stored in-memory. This enables data-driven selection decisions based on observed model behavior (latency, error rates, rate limit status) rather than static configuration alone.
```

ronenkat (approved):

nirrozenbaum:
PR approved by @ronenkat. 
this version is good enough for start moving forward and evolve in iterations.

PR #58, opened 2026-05-10
skeleton for model selector profile with plugins

nirrozenbaum:
## What does this PR do?

initial skeleton for model selection types and structure. 

## How was this tested?

## Checklist

### DIFF
--- a/pkg/framework/modelselector/plugins.go
+++ b/pkg/framework/modelselector/plugins.go
@@ -30,7 +30,7 @@ type Filter interface {
 }
 
 // Scorer defines the interface for scoring a list of models based on context.
-// Scorers must score pods with a value within the range of [0,1] where 1 is the highest score.
+// Scorers must score models with a value within the range of [0,1] where 1 is the highest score.
 // If a scorer returns value greater than 1, it will be treated as score 1.
 // If a scorer returns value lower than 0, it will be treated as score 0.
 type Scorer interface {
--- a/pkg/framework/modelselector/plugins.go
+++ b/pkg/framework/modelselector/plugins.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package modelselector
+
+import (
+	"context"
+
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework"
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/datalayer"
+)
+
+// Filter defines the interface for filtering a list of candidate models based on context.
+type Filter interface {
+	framework.Plugin
+	Filter(ctx context.Context, cycleState *framework.CycleState, request *framework.InferenceRequest, models []datalayer.Model) []datalayer.Model
+}
+
+// Scorer defines the interface for scoring a list of models based on context.
+// Scorers must score pods with a value within the range of [0,1] where 1 is the highest score.
+// If a scorer returns value greater than 1, it will be treated as score 1.
+// If a scorer returns value lower than 0, it will be treated as score 0.
+type Scorer interface {
+	framework.Plugin
+	Score(ctx context.Context, cycleState *framework.CycleState, request *framework.InferenceRequest, models []datalayer.Model) map[datalayer.Model]float64
+}
+
+// Picker picks the final model(s) to send the request to.
+type Picker interface {
+	framework.Plugin
+	Pick(ctx context.Context, cycleState *framework.CycleState, scoredModels []*ScoredModel) *ProfileRunResult
+}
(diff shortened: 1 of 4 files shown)


pkg/modelselector/model_selector_profile.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// ModelSelectorProfile provides a profile configuration for the model-selector which influence model decisions.
+type ModelSelectorProfile struct {
+	filters []modelselector.Filter
+	scorers []*WeightedScorer
+	picker  modelselector.Picker
+}
+
+func (p *ModelSelectorProfile) String() string {
```
nirrozenbaum, line 42: can be used on startup to log the ModelSelectorProfile. 
all the logic should be pushed in a follow up

pkg/framework/modelselector/plugins.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// Filter defines the interface for filtering a list of candidate models based on context.
+type Filter interface {
+	framework.Plugin
+	Filter(ctx context.Context, cycleState *framework.CycleState, request *framework.InferenceRequest, models []datalayer.Model) []datalayer.Model
+}
+
+// Scorer defines the interface for scoring a list of models based on context.
+// Scorers must score pods with a value within the range of [0,1] where 1 is the highest score.
```
Mohammad-nassar10:
```suggestion
// Scorers must score models with a value within the range of [0,1] where 1 is the highest score.
```

pkg/modelselector/model_selector_profile.go
```diff
@@ -0,0 +1,58 @@
...
+package modelselector
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/modelselector"
+)
+
+// NewSchedulerProfile creates a new SchedulerProfile object and returns its pointer.
```
Mohammad-nassar10:
```suggestion
// NewModelSelectorProfile creates a new ModelSelectorProfile object and returns its pointer.
```

pkg/modelselector/model_selector_profile.go
```diff
@@ -0,0 +1,58 @@
...
+func NewModelSelectorProfile() *ModelSelectorProfile {
+	return &ModelSelectorProfile{
+		filters: []modelselector.Filter{},
+		scorers: []*WeightedScorer{},
+		// picker remains nil since profile doesn't support multiple pickers
+	}
+}
+
+// ModelSelectorProfile provides a profile configuration for the model-selector which influence model decisions.
+type ModelSelectorProfile struct {
```
Mohammad-nassar10, line 36: Should the `Run()` be implemented to satisfy the `ModelSelectorProfile` interface?

pkg/framework/modelselector/plugins.go
noyitz: nit: copy-paste from upstream — should say "models" instead of "pods"
```suggestion
// Scorers must score models with a value within the range of [0,1] where 1 is the highest score.
```

pkg/modelselector/model_selector_profile.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/modelselector"
+)
+
+// NewSchedulerProfile creates a new SchedulerProfile object and returns its pointer.
+func NewModelSelectorProfile() *ModelSelectorProfile {
+	return &ModelSelectorProfile{
+		filters: []modelselector.Filter{},
```
noyitz, line 29: nit: copy-paste from upstream — comment references SchedulerProfile instead of ModelSelectorProfile
```suggestion
// NewModelSelectorProfile creates a new ModelSelectorProfile object and returns its pointer.
```

pkg/framework/modelselector/types.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/datalayer"
+)
+
+type ScoredModel struct {
+	datalayer.Model
+	Score float64
+}
+
+// ProfileRunResult captures the profile run result.
+type ProfileRunResult struct {
```
noyitz, line 32: The proposal states "Picker MUST return one model at minimum" — should this be a single model rather than a slice? The current design allows returning multiple models which contradicts the proposal and could create ambiguity about which model to actually use downstream.
```suggestion
	TargetModel datalayer.Model
```

pkg/modelselector/model_selector_profile.go
nirrozenbaum: sure :). 
the intention in this PR was to only add the types to allow @noyitz to work on the profile and framework logic, while others can work in parallel on scorers/filters. Run (and the extra logic) should be added in follow ups.

pkg/framework/modelselector/plugins.go
nirrozenbaum: fixed

nirrozenbaum: fixed

pkg/framework/modelselector/types.go
nirrozenbaum: yes. we should start with a single model. 
fixed

pkg/modelselector/model_selector_profile.go
nirrozenbaum: fixed

pkg/modelselector/model_selector_profile.go
nirrozenbaum: fixed

ronenkat (approved):

Issue #50, opened 2026-05-07
[Feature]: Data Layer for background metrics tracking


Mohammad-nassar10:
### Problem Statement

IPP currently has no mechanism to track per-request runtime metrics (in-flight counts, latency)
that the Model Selector can use to make smarter routing decisions.

This is a prerequisite for issue #42.

### Proposed Solution

Define a Data Layer - an async background pipeline that collects request/response events
off the critical path, processes them through pluggable Extractors, and writes aggregates
to the `DataStore` for the Model Selector to consume.

### Willingness to Contribute

Yes, I can submit a PR

PR #49, opened 2026-05-07
Data Layer proposal.

Mohammad-nassar10:
## What does this PR do?
Adds a design proposal for the IPP Data Layer (`docs/proposals/data-layer-proposal/README.md`).
Defines a background tracking pipeline that runs outside the critical request path to collect and calculate different attributes and metrics.

## Why is this change needed?
The data layer provides the foundation for tracking per-model metrics without adding latency to the request hot path.

## How was this tested?

- [ ] Unit tests added/updated
- [ ] Integration/e2e tests added/updated
- [ ] Manual testing performed

## Checklist

- [ ] Commits are signed off (`git commit -s`) per [DCO](PR_SIGNOFF.md)
- [ ] Code follows project [contributing guidelines](CONTRIBUTING.md)
- [ ] Tests pass locally (`make test`)
- [ ] Linters pass (`make lint`)
- [ ] Documentation updated (if applicable)

## Related Issues
Related to issue #50

PR #60, opened 2026-05-10
Add data layer interfaces.

Mohammad-nassar10:
## What does this PR do?

Adds the framework-level contracts for the asynchronous tracking pipeline: `EventNotifier` for non-blocking event dispatch, `Extractor` for batch processing, and `NotificationSource` to compose them into a managed pipeline.

## How was this tested?

## Checklist

## Related Issues
Related to #50.


pkg/framework/datalayer.go
```diff
@@ -0,0 +1,84 @@
...
+// Event is the carrier for all data layer events.
+type Event struct {
+	Type    EventType
+	Payload any
+}
+
+// RequestPayload is the Payload for RequestEventType.
+type RequestPayload struct {
+	Request   *InferenceRequest
+	StartTime time.Time
```
ronenkat: What does the StartTime refer to?

pkg/framework/datalayer.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// ResponsePayload is the Payload for ResponseEventType.
+type ResponsePayload struct {
+	Request  *InferenceRequest
+	Response *InferenceResponse
+	Duration time.Duration
```
ronenkat: What would be the duration? how would we translate it to TTFT or end-to-end time?

pkg/framework/datalayer.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"context"
+	"time"
+
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/datalayer"
+)
+
+// DataStore provides model-keyed access to aggregated runtime metrics.
+type DataStore interface {
```
ronenkat: What is the difference versus `type Datastore interface` in #34

pkg/framework/datalayer.go
Mohammad-nassar10: The timestamp when the server first sees the request. It marks the start of the request lifecycle.

pkg/framework/datalayer.go
Mohammad-nassar10: This is end-to-end latency from when request headers arrived to when the full response body was received.
To measure TTFT we would need to record a separate timestamp when the first response body chunk arrives in the streaming path, the current implementation accumulates all chunks before `HandleResponseBody` is called. 
Will handle the TTFT in separate PR (maybe adding a new event for this attribute).

pkg/framework/datalayer.go
Mohammad-nassar10: These are intentionally different in scope. `datastore.Datastore` (PR #34) is the full management interface (`GetOrCreateModel` + `DeleteModel` + `Models()`) used by the runner to own the store lifecycle. `framework.DataStore` here is a narrower interface with only `GetOrCreateModel` that follows interface segregation: plugins get only what they need and cannot delete or enumerate the full model registry.

nirrozenbaum (commented):
@Mohammad-nassar10 can you please the intention of the PR?
in epp - data source used to collect data for all endpoints and data extractor is used to extract subset of the collected data. 
as far as I understand - in the first iteration we plan to calculate metrics on model from IPP perspective, e.g., calculate TTFT inside IPP and store that information. 

do you have an example for datasource/extractor that you had in mind?

pkg/framework/datalayer.go
nirrozenbaum: I think that should connect to the code here (and not in framework):
https://github.com/llm-d/llm-d-inference-payload-processor/blob/9eb517b89360da9cef7e487324065b31534cea30/pkg/handlers/server.go#L64

previous PRs already prepared the ground for calculating TTFT and other latencies measurements, but this is not used yet.

Mohammad-nassar10:
@nirrozenbaum this introduces a background tracking pipeline that runs outside the critical request path. The intent is rather than computing metrics synchronously on every request, the handlers fire lightweight non-blocking events, and a background goroutine batches and dispatches them to Extractors every specific time.

An example of an extractor is `RunningRequestsExtractor` which on each `RequestEvent` increments in-flight request and on each `ResponseEvent` decrements them (initial extractor). It writes a `RunningRequestCount` for each model in the datastore.

More details can be found in [this proposal document](https://github.com/llm-d/llm-d-inference-payload-processor/pull/49/changes).

pkg/framework/datalayer.go
Mohammad-nassar10: The payload carries these measurements/info so extractors can compute duration/latency/TTFT in the background and updating the datastore.

ronenkat: lets use the same terms as in RequestContext
check if you need startTime in the request event - as you can send it in the response event

Mohammad-nassar10: Deleted it, Not really needed, we got the duration in the response.

pkg/framework/datalayer.go
```diff
@@ -0,0 +1,83 @@
...
+// RequestPayload is the Payload for RequestEventType.
+type RequestPayload struct {
+	Request *InferenceRequest
+}
+
+// ResponsePayload is the Payload for ResponseEventType.
+type ResponsePayload struct {
+	Request  *InferenceRequest
+	Response *InferenceResponse
+	Duration time.Duration
```
ronenkat, line 62: i don't think you need Duration as you should have `ResponseCompleteTimestamp`

Mohammad-nassar10: we need either the `Duration` or to get both the `ResponseCompleteTimestamp` and `RequestReceivedTimestamp` in the payload. We can keep Duration for now and revisit in the wiring PR (between server and datalayer) where the actual computation will live.

ronenkat (approved):

pkg/framework/datalayer.go
```diff
@@ -0,0 +1,83 @@
...
+// DataStore provides model-keyed access to aggregated runtime metrics.
+type DataStore interface {
+	GetOrCreateModel(name string) datalayer.Model
+}
+
+// DataSource is the base interface for background data layer components.
+type DataSource interface {
+	Plugin
+	Start(ctx context.Context) error
+	Stop()
```
shmuelk, line 36: Do you really need a Stop() function? Won't simply canceling the Context passed to Start() do the trick?

shmuelk:
I think you will need to benchmark just much "cheaper" this background processing really is. How much processing will the Extractors actually do?

You will need some sort of thread safe queueing mechanism in any event.

pkg/framework/datalayer.go
Mohammad-nassar10: Canceling the context tells the goroutine "please stop soon", but returns immediately (the goroutine may still be mid-batch). But, `Stop()` waits until it's actually done. This is better for clean shutdown and for tests that need to check results after stopping.

shmuelk: You should at least document that behavior of Stop(). Sometimes a Stop() function simply does a cancel on a context somewhere or other.

Mohammad-nassar10:
@shmuelk Thanks! agree this is worth measuring. The main benefit of the background model is decoupling: request latency is never affected by extraction work, regardless of how slow an extractor becomes.
Another benefit is saving `Datastore` writes to one write per tick instead on each event.
Also, in the next steps we have a plan to fetch metrics from EPPs which involves network I/O that add significant latency. So I think it is worth it to add the background processing.

Regarding the queuing mechanism we are using `ch chan Event` which is thread safe.

pkg/framework/datalayer.go
Mohammad-nassar10: Added a doc comment for clarification. Thanks!

shmuelk (approved):

Mohammad-nassar10:
Rebased to fix the previous datastore lint error.

PR #62, opened 2026-05-11
fix(build): improve image-build target for local development

yehuditkerido:
## What does this PR do?

Replace `docker buildx` with standard `docker build` in `image-build` target and pass version info (`COMMIT_SHA`, `BUILD_REF`) to the Dockerfile. Adds support for Podman via `CONTAINER_RUNTIME` variable.

## Why is this change needed?

The previous `image-build` target failed locally with "Multi-platform build is not supported for the docker driver". This aligns the target with `llm-d-inference-scheduler` patterns and properly embeds version info in the binary.


nirrozenbaum:
/lgtm
/approve

Thanks!

PR #72, opened 2026-05-11
feat: implement model selector pipeline

### LINKED ISSUE #65: implement model selector pipeline
we should implement `ModelSelector` (in /pkg/modelselector/modelselector.go file).
the implementation should drive inspiration from [Scheduler](https://github.com/llm-d/llm-d-inference-scheduler/blob/434575493d5bf3a5c7a5b9b1d6210ca6228ce076/pkg/epp/scheduling/scheduler.go#L48) and include a single Public function `SelectModel` or just `Select`. 
in the initial version - ModelSelector can assume a single profile (no need to support ProfilePicker).

this should also include ModelSelectorProfile.Run function implementation and adding metrics reporting latencies of each plugin.

/cc @noyitz

noyitz:
## What does this PR do?

Implements the ModelSelector pipeline with a single public `Select` method and the `ModelSelectorProfile.Run` execution logic (Filter → Score → Pick), following the upstream Scheduler pattern from [llm-d-inference-scheduler](https://github.com/llm-d/llm-d-inference-scheduler/blob/main/pkg/epp/scheduling/scheduler.go).

### New files:
- `pkg/modelselector/modelselector.go` — `ModelSelector` struct with `Select` method
- `pkg/modelselector/modelselector_test.go` — tests for ModelSelector
- `pkg/modelselector/model_selector_profile_test.go` — tests for profile pipeline

### Modified files:
- `pkg/modelselector/model_selector_profile.go` — added `Run()`, builder methods (`WithFilters`, `WithScorers`, `WithPicker`, `AddPlugins`), `enforceScoreRange`
- `pkg/metrics/metrics.go` — added model selector E2E latency and attempt metrics

### Key design decisions:
- Single profile (no ProfileHandler/ProfilePicker) per issue requirements
- Score range enforcement [0,1] matching upstream pattern
- Per-plugin latency metrics using existing `RecordPluginProcessingLatency` with ModelSelector-specific extension points
- E2E latency and success/failure attempt metrics

## Why is this change needed?

Implements issue #65 — the core model selection pipeline that all future filter/scorer/picker plugins will plug into.

## How was this tested?

- [x] Unit tests added/updated

15 tests covering:
- Full pipeline execution (filter → score → pick)
- Filter execution order (sequential, each receives previous output)
- Score weight accumulation across multiple scorers
- Score range enforcement (clamping to [0,1])
- AddPlugins validation (scorer without weight, duplicate picker)
- Empty/nil candidate handling
- End-to-end selection with filter + scorer + picker

## Checklist

## Related Issues

Closes #65

### DIFF
--- a/pkg/modelselector/model_selector_profile.go
+++ b/pkg/modelselector/model_selector_profile.go
@@ -140,10 +140,6 @@ func (p *ModelSelectorProfile) String() string {
 
 // Run runs the ModelSelectorProfile pipeline: Filter → Score → Pick.
 func (p *ModelSelectorProfile) Run(ctx context.Context, request *framework.InferenceRequest, cycleState *framework.CycleState, candidateModels []datalayer.Model) (*modelselector.ProfileRunResult, error) {
-	if p.picker == nil {
-		return nil, errors.New("no picker plugin configured")
-	}
-
 	models := p.runFilterPlugins(ctx, request, cycleState, candidateModels)
 	if len(models) == 0 {
 		return nil, errors.New("no models available after filtering")
@@ -152,9 +148,6 @@ func (p *ModelSelectorProfile) Run(ctx context.Context, request *framework.Infer
 	weightedScorePerModel := p.runScorerPlugins(ctx, request, cycleState, models)
 
 	result := p.runPickerPlugin(ctx, cycleState, weightedScorePerModel)
-	if result == nil || result.TargetModel == nil {
-		return nil, errors.New("picker returned no result")
-	}
 
 	return result, nil
 }
@@ -180,12 +173,12 @@ func (p *ModelSelectorProfile) runFilterPlugins(ctx context.Context, request *fr
 	return filteredModels
 }
 
-func (p *ModelSelectorProfile) runScorerPlugins(ctx context.Context, request *framework.InferenceRequest, cycleState *framework.CycleState, models []datalayer.Model) map[string]scoredModelEntry {
+func (p *ModelSelectorProfile) runScorerPlugins(ctx context.Context, request *framework.InferenceRequest, cycleState *framework.CycleState, models []datalayer.Model) map[string]*modelselector.ScoredModel {
 	logger := log.FromContext(ctx)
 
-	entries := make(map[string]scoredModelEntry, len(models))
+	scoredModels := make(map[string]*modelselector.ScoredModel, len(models))
 	for _, model := range models {
-		entries[model.GetName()] = scoredModelEntry{model: model, score: 0}
+		scoredModels[model.GetName()] = &modelselector.ScoredModel{Model: model, Score: 0}
 	}
 
 	for _, scorer := range p.scorers {
@@ -194,29 +187,25 @@ func (p *ModelSelectorProfile) runScorerPlugins(ctx context.Context, request *fr
 		scores := scorer.Score(ctx, cycleState, request, models)
 		metrics.RecordPluginProcessingLatency(scorerExtensionPoint, scorer.TypedName().Type, scorer.TypedName().Name, time.Since(before))
 		for model, score := range scores {
-			if entry, exists := entries[model.GetName()]; exists {
-				entry.score += enforceScoreRange(score) * scorer.Weight()
-				entries[model.GetName()] = entry
+			if sm, exists := scoredModels[model.GetName()]; exists {
+				sm.Score += enforceScoreRange(score) * scorer.Weight()
 			}
 		}
 		logger.V(logutil.DEBUG).Info("Completed running scorer plugin", "plugin", scorer.TypedName())
 	}
 	logger.V(logutil.VERBOSE).Info("Completed running scorer plugins")
 
-	return entries
+	return scoredModels
 }
 
-type scoredModelEntry struct {
-	model datalayer.Model
-	score float64
-}
-
-func (p *ModelSelectorProfile) runPickerPlugin(ctx context.Context, cycleState *framework.CycleState, entries map[string]scoredModelEntry) *modelselector.ProfileRunResult {
+func (p *ModelSelectorProfile) runPickerPlugin(ctx context.Context, cycleState *framework.CycleState, scoredModelMap map[string]*modelselector.ScoredModel) *modelselector.ProfileRunResult {
 	logger := log.FromContext(ctx)
 
-	scoredModels := make([]*modelselector.ScoredModel, 0, len(entries))
-	for _, entry := range entries {
-		scoredModels = append(scoredModels, &modelselector.ScoredModel{Model: entry.model, Score: entry.score})
+	scoredModels := make([]*modelselector.ScoredModel, len(scoredModelMap))
+	i := 0
+	for _, sm := range scoredModelMap {
+		scoredModels[i] = sm
+		i++
 	}
 
 	logger.V(logutil.VERBOSE).Info("Running picker plugin", "plugin", p.picker.TypedName())
--- a/pkg/modelselector/model_selector_profile.go
+++ b/pkg/modelselector/model_selector_profile.go
@@ -31,6 +31,9 @@ import (
 	"github.com/llm-d/llm-d-inference-payload-processor/pkg/metrics"
 )
 
+// compile-time interface validation
+var _ modelselector.ModelSelectorProfile = &ModelSelectorProfile{}
+
 const (
 	filterExtensionPoint = "ModelSelectorFilter"
 	scorerExtensionPoint = "ModelSelectorScorer"
@@ -105,16 +108,25 @@ func (p *ModelSelectorProfile) String() string {
 		scorerNames[i] = fmt.Sprintf("%s: %f", scorer.TypedName(), scorer.Weight())
 	}
 
+	pickerName := "<none>"
+	if p.picker != nil {
+		pickerName = p.picker.TypedName().String()
+	}
+
 	return fmt.Sprintf(
 		"{Filters: [%s], Scorers: [%s], Picker: %s}",
 		strings.Join(filterNames, ", "),
 		strings.Join(scorerNames, ", "),
-		p.picker.TypedName(),
+		pickerName,
 	)
 }
 
 // Run runs the ModelSelectorProfile pipeline: Filter → Score → Pick.
 func (p *ModelSelectorProfile) Run(ctx context.Context, request *framework.InferenceRequest, cycleState *framework.CycleState, candidateModels []datalayer.Model) (*modelselector.ProfileRunResult, error) {
+	if p.picker == nil {
+		return nil, fmt.Errorf("no picker plugin configured")
+	}
+
 	models := p.runFilterPlugins(ctx, request, cycleState, candidateModels)
 	if len(models) == 0 {
 		return nil, fmt.Errorf("no models available after filtering")
@@ -162,7 +174,9 @@ func (p *ModelSelectorProfile) runScorerPlugins(ctx context.Context, request *fr
 		scores := scorer.Score(ctx, cycleState, request, models)
 		metrics.RecordPluginProcessingLatency(scorerExtensionPoint, scorer.TypedName().Type, scorer.TypedName().Name, time.Since(before))
 		for model, score := range scores {
-			weightedScorePerModel[model] += enforceScoreRange(score) * scorer.Weight()
+			if _, exists := weightedScorePerModel[model]; exists {
+				weightedScorePerModel[model] += enforceScoreRange(score) * scorer.Weight()
+			}
 		}
 		logger.V(logutil.DEBUG).Info("Completed running scorer plugin", "plugin", scorer.TypedName())
 	}
--- a/pkg/modelselector/model_selector_profile.go
+++ b/pkg/modelselector/model_selector_profile.go
@@ -79,22 +79,38 @@ func (p *ModelSelectorProfile) WithPicker(picker modelselector.Picker) *ModelSel
 // Special Case: In order to add a scorer, one must use NewWeightedScorer in order to provide a weight.
 // If a scorer implements more than one interface, supplying a WeightedScorer is sufficient.
 func (p *ModelSelectorProfile) AddPlugins(pluginObjects ...framework.Plugin) error {
-	for _, plugin := range pluginObjects {
-		if weightedScorer, ok := plugin.(*WeightedScorer); ok {
-			p.scorers = append(p.scorers, weightedScorer)
-			plugin = weightedScorer.Scorer
-		} else if scorer, ok := plugin.(modelselector.Scorer); ok {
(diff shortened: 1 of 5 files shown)


pkg/modelselector/modelselector.go
```diff
@@ -0,0 +1,72 @@
...
+		metrics.RecordModelSelectorE2ELatency(time.Since(selectStart))
+		metrics.RecordModelSelectorAttempt(err)
+	}()
+
+	if len(candidateModels) == 0 {
+		err = fmt.Errorf("no candidate models provided")
+		return nil, err
+	}
+
+	cycleState := framework.NewCycleState()
```
nirrozenbaum: cycle state would be received from outside (shared with other payload processing plugins.

```suggestion
```

pkg/modelselector/model_selector_profile.go
```diff
@@ -56,3 +112,88 @@ func (p *ModelSelectorProfile) String() string {
...
+		before := time.Now()
+		filteredModels = filter.Filter(ctx, cycleState, request, filteredModels)
+		metrics.RecordPluginProcessingLatency(filterExtensionPoint, filter.TypedName().Type, filter.TypedName().Name, time.Since(before))
+		logger.V(logutil.DEBUG).Info("Completed running filter plugin", "plugin", filter.TypedName(), "remainingModels", len(filteredModels))
+		if len(filteredModels) == 0 {
+			logger.V(logutil.VERBOSE).Info("Filter eliminated all models", "plugin", filter.TypedName())
+			break
+		}
+	}
+	logger.V(logutil.VERBOSE).Info("Completed running filter plugins", "remainingModels", len(filteredModels))
```
nirrozenbaum: nit: can we remove the remaining models from the completion log?
it's logged in prev log line (when completing last filter).
 
```suggestion
	logger.V(logutil.VERBOSE).Info("Completed running filter plugins")
```

pkg/modelselector/model_selector_profile_test.go
```diff
@@ -0,0 +1,354 @@
...
+			},
+			wantModel: "model-c",
+		},
+		{
+			name:   "all models filtered returns error",
+			models: []datalayer.Model{modelA, modelB},
+			filter: &testFilter{
+				typedName: framework.TypedName{Type: "test-filter", Name: "block-all"},
+				filterFn: func(_ []datalayer.Model) []datalayer.Model {
+					return nil
```
nirrozenbaum: maybe it's better to return []datalayer.Model{} // empty slice, not nil.
or does filter function return nil when all models are filtered?

pkg/modelselector/model_selector_profile_test.go
```diff
@@ -0,0 +1,354 @@
...
+				},
+			},
+			wantErr: true,
+		},
+		{
+			name:      "no scorers gives equal scores",
+			models:    []datalayer.Model{modelA, modelB},
+			wantModel: "", // any model is fine with equal scores
```
nirrozenbaum: what are we testing here?

pkg/modelselector/model_selector_profile_test.go
```diff
@@ -0,0 +1,354 @@
...
+				t.Fatal("expected result, got nil")
+			}
+			if tt.wantModel != "" && result.TargetModel.GetName() != tt.wantModel {
+				t.Errorf("expected model %q, got %q", tt.wantModel, result.TargetModel.GetName())
+			}
+		})
+	}
+}
+
+func TestFilterExecutionOrder(t *testing.T) {
```
nirrozenbaum: I think this one can be removed. there is no meaning for the filters execution order (it's just a set of predicates and all of them should run).

pkg/modelselector/model_selector_profile_test.go
```diff
@@ -0,0 +1,354 @@
...
+					scores[m] = 0.0
+				} else {
+					scores[m] = 1.0
+				}
+			}
+			return scores
+		},
+	}
+
+	// cost weight=3, latency weight=1 → model-a should win (3*1.0 + 1*0.0 = 3.0 vs 3*0.0 + 1*1.0 = 1.0)
```
nirrozenbaum, line 233: nice!

noyitz:
@nirrozenbaum ready for review

pkg/modelselector/modelselector.go
noyitz: Done. Select() now accepts CycleState as a parameter.

pkg/modelselector/model_selector_profile.go
noyitz: Done.

pkg/modelselector/model_selector_profile_test.go
noyitz: Good point. Changed to return empty slice.

pkg/modelselector/model_selector_profile_test.go
noyitz: Renamed to "picker runs with zero scores when no scorers configured".

pkg/modelselector/model_selector_profile_test.go
noyitz: Agreed. Removed.

pkg/modelselector/model_selector_profile.go
```diff
@@ -39,6 +55,49 @@ type ModelSelectorProfile struct {
...
+	p.picker = picker
+	return p
+}
+
+// AddPlugins adds the given plugins to the profile according to the interfaces each plugin implements.
+// A plugin may implement more than one interface.
+// Special Case: In order to add a scorer, one must use NewWeightedScorer in order to provide a weight.
+// If a scorer implements more than one interface, supplying a WeightedScorer is sufficient.
+func (p *ModelSelectorProfile) AddPlugins(pluginObjects ...framework.Plugin) error {
+	for _, plugin := range pluginObjects {
```
shaneutt: The internal state of the MSP can be modified during this loop, but an error can also occur. If both of those happen then the MSP can be in an inconsistent state.

You could validate _then_ apply to avoid ending up with a defunct profile.

pkg/modelselector/model_selector_profile.go
```diff
@@ -49,10 +108,106 @@ func (p *ModelSelectorProfile) String() string {
...
+	models := p.runFilterPlugins(ctx, request, cycleState, candidateModels)
+	if len(models) == 0 {
+		return nil, fmt.Errorf("no models available after filtering")
+	}
+
+	weightedScorePerModel := p.runScorerPlugins(ctx, request, cycleState, models)
+
+	result := p.runPickerPlugin(ctx, cycleState, weightedScorePerModel)
+
+	return result, nil
```
shaneutt, line 152: Technically if result is ever `nil` then you would return (`nil`, `nil`) here and it would look like a successful result, but could lead to nil pointer dereferences.

Perhaps `runPickerPlugin` would be better to return an error in case it ever manages to produce no result?

pkg/modelselector/model_selector_profile.go
```diff
@@ -49,10 +108,106 @@ func (p *ModelSelectorProfile) String() string {
...
+	}
+
+	for _, scorer := range p.scorers {
+		logger.V(logutil.VERBOSE).Info("Running scorer plugin", "plugin", scorer.TypedName())
+		before := time.Now()
+		scores := scorer.Score(ctx, cycleState, request, models)
+		metrics.RecordPluginProcessingLatency(scorerExtensionPoint, scorer.TypedName().Type, scorer.TypedName().Name, time.Since(before))
+		for model, score := range scores {
+			if _, exists := weightedScorePerModel[model]; exists {
+				weightedScorePerModel[model] += enforceScoreRange(score) * scorer.Weight()
```
shaneutt: Using the model as the key here means equality check will be on pointer identity. This might lead to some unintended consequences? For instance, if a scorer implementation creates a new Model instead rather than reusing the instances it received, all its scores would be silently ignored.

Might be better to make a more explicit contract and validation around this now, to avoid some potential "fun" debugging later? 🤔

nirrozenbaum:
@noyitz can fix linter issues?

pkg/modelselector/model_selector_profile.go
nirrozenbaum: AddPlugins is called only during bootstrap cause we don't support dynamic addition of plugins.. 
so if an error is caught during startup, we log it and exit...

noyitz: Good catch. Refactored to validate-then-apply — all plugins are validated first, state is only modified after all pass.

pkg/modelselector/model_selector_profile.go
noyitz: Added nil check after runPickerPlugin — returns an error if picker produces no result.

pkg/modelselector/model_selector_profile.go
noyitz: Great point. Changed the score map to use model name (string) as key instead of the Model interface. This avoids the pointer identity issue — scorers that return new Model instances with the same name will still match correctly.

noyitz:
@nirrozenbaum lint issue is fixed — replaced all `fmt.Errorf` with `errors.New` where no format verbs are used (the 4 perfsprint findings). Also addressed Shane's 3 review comments. CI needs maintainer approval to re-run on the fork.

noyitz:
@nirrozenbaum all review comments addressed (yours and Shane's). Ready for re-review when you get a chance.

pkg/modelselector/modelselector.go
```diff
@@ -0,0 +1,75 @@
...
+	result, err = s.profile.Run(ctx, request, cycleState, candidateModels)
+	if err != nil {
+		logger.V(logutil.VERBOSE).Info("Model selection failed", "error", err.Error())
+		return nil, err
+	}
+
+	if result == nil || result.TargetModel == nil {
+		err = errors.New("model selection returned no result")
+		return nil, err
+	}
```
Mohammad-nassar10, line 70: `Run()` already guarantees non-nil result or returns an error, so the check here is unreachable.

```suggestion
```

pkg/modelselector/model_selector_profile.go
```diff
@@ -49,10 +125,115 @@ func (p *ModelSelectorProfile) String() string {
...
-		p.picker.TypedName(),
+		pickerName,
 	)
 }
+
+// Run runs the ModelSelectorProfile pipeline: Filter → Score → Pick.
+func (p *ModelSelectorProfile) Run(ctx context.Context, request *framework.InferenceRequest, cycleState *framework.CycleState, candidateModels []datalayer.Model) (*modelselector.ProfileRunResult, error) {
+	if p.picker == nil {
+		return nil, errors.New("no picker plugin configured")
+	}
```
nirrozenbaum: nit: this is not an error in the data plane. this is caught during bootstrap, if no picker is configured, IPP should log appropriate message and exit. 

the comment is not about this specific check which is 2 lines, but more about the state of mind - we shouldn't add complexity to the code and checks if it's not relevant. if picker is validated once at startup, we can assume it's valid (plugins are configured statically, only at startup).

```suggestion
```

pkg/modelselector/model_selector_profile.go
```diff
@@ -49,10 +125,115 @@ func (p *ModelSelectorProfile) String() string {
...
+	logger.V(logutil.VERBOSE).Info("Completed running scorer plugins")
+
+	return entries
+}
+
+type scoredModelEntry struct {
+	model datalayer.Model
+	score float64
+}
```
nirrozenbaum: why do we have this type instead of reusing ScoredModel from framework dir?
https://github.com/llm-d/llm-d-inference-payload-processor/blob/fe26f25b0f6ffa79f740b924deea6e3d2ed03fad/pkg/framework/modelselector/types.go#L26-L29

pkg/modelselector/model_selector_profile.go
```diff
@@ -49,10 +125,115 @@ func (p *ModelSelectorProfile) String() string {
...
+
+func (p *ModelSelectorProfile) runPickerPlugin(ctx context.Context, cycleState *framework.CycleState, entries map[string]scoredModelEntry) *modelselector.ProfileRunResult {
+	logger := log.FromContext(ctx)
+
+	scoredModels := make([]*modelselector.ScoredModel, 0, len(entries))
+	for _, entry := range entries {
+		scoredModels = append(scoredModels, &modelselector.ScoredModel{Model: entry.model, Score: entry.score})
+	}
```
nirrozenbaum: append operations in this context are wasteful. we can easily use index for O(1) operations:
see example for schduler:
https://github.com/llm-d/llm-d-inference-scheduler/blob/3fd8146a277b54e42bde39e9a728646aa8d9b21a/pkg/epp/scheduling/scheduler_profile.go#L178-L183

pkg/modelselector/modelselector.go
nirrozenbaum: agreed. the check is duplicated. 
I would personally leave the check here and remove it from profile.

pkg/modelselector/model_selector_profile.go
```diff
@@ -39,6 +56,65 @@ type ModelSelectorProfile struct {
...
+				if newPicker != nil {
+					existing = newPicker
+				}
+				return fmt.Errorf("failed to set '%s' as picker, already have a registered picker plugin '%s'", picker.TypedName(), existing.TypedName())
+			}
+			newPicker = picker
+		}
+	}
+
+	// Apply after successful validation
```
ronenkat, line 109: should we check for duplicate plugins? i.e., adding again the same plugin.

pkg/modelselector/modelselector.go
noyitz: Done. Removed the nil result check from Run() — keeping it only in Select() per Nir's suggestion.

pkg/modelselector/model_selector_profile.go
noyitz: Agreed. Removed — picker validation belongs at startup, not on every request.

pkg/modelselector/model_selector_profile.go
noyitz: Good catch. Removed scoredModelEntry, now reusing modelselector.ScoredModel directly.

pkg/modelselector/model_selector_profile.go
noyitz: Done. Switched to indexed assignment.

pkg/modelselector/modelselector.go
noyitz: Done. Check stays in Select(), removed from Run().

pkg/modelselector/model_selector_profile.go
noyitz: The upstream scheduler doesn't check for duplicates either. I'd keep it simple for now and add duplicate detection as a future enhancement if it becomes an issue in practice.

Issue #77, opened 2026-05-12
create plugins config file + loader for pluggability


nirrozenbaum:
build config file loader for payload processor plugins.
conceptually it should behave similarly to epp config loader other than few differences.

in terms of structure, it should be a yaml, CRD style (but not a CRD) and may look like the following:
```yaml
apiVersion: inference.llm-d.ai/v1alpha1
kind: PayloadProcessorConfig 
plugins:
  - type: <plugin-type1>
    name: (optional plugin name, if not specified, defaults in code to plugin type)
  - type: <plugin-type2> (could be either request processor plugin, response processor plugin, or model selector plugin which is from type filter/scorer/picker)
   ......
  - type: <profile-picker> // we must have exactly one profile picker 
modelSelector:
  - pluginRef: rate-limit-filter
  - pluginRef: cost-scorer
    weight: 2
  - pluginRef: latency-scorer
    weight: 3
  - pluginRef: max-score-picker // must have exactly one picker
profiles:
  - name: <profile-name>
    request: 
      - pluginRef: <plugin1 name>
        parameters:
                ...
                ...
      - <HERE WE NEED TO BE ABLE TO SET MODEL SELECTOR>
      - pluginRef: <plugin1 name>
        parameters:
                ...
                ...
    response:
      - pluginRef: <plugin2 name>
        parameters:
                ...
                ...
  - name: <profile-name2>
                ...
                ...
``` 

key differences from epp config:
- we introduce the concept of model selector (which has filter/scorer/picker plugins). is should be possible to define the order of execution of model selector along with other request processor plugins (only on the request, model selector is not valid on response). for example, define request guard first (request processor), then model selector, and then additional request processor.
- we want to introduce the concept of different `profile`(collection of plugins) per request. that may depend on request properties (e.g., does it contain model in the body or not, if not we want to run the profile with model selector). so we need to be able to have multiple profiles and profile picker. we may want to explore the option to include profile A in profile B (e.g., profile B need to do profile A plus one more plugin before/after). including of profile can be deferred to later stage.
- In the initial version we can be explicit and don't use default values. config loader can fail if the configuration is wrong. we can iterate on this point.
- as we evolve we might need to add the feature gates section but that's not needed for the inital impl.
- this is just initial thinking, nothing is written in stone, so feel free to propose difference ways to represent the information. 

This is a big task that could be split to multiple sub tasks (e.g., defining the CRD style api, loading raw config, adding defaults, unit tests, etc).

[...]

PR #80, opened 2026-05-12
Add initial configuration structs with code generation machinery

shmuelk:
## What does this PR do?
This PR is the first in series of PRs that add a textual configuration to the Inference Payload Processor(IPP).

In particular it adds an initial configuration struct, the generated code, and the needed build machinery.

## Why is this change needed?
There is a desire to configure various parts of the IPP from a configuration file rather than command line arguments as is done today.

The configuration will be YAML text that looks like a Kubernetes (K8S) Custom Resource (CR), but it isn't one. The K8S infrastructure is used for reading the YAML text and loading it into a struct as well as generating some of the helper code.

## How was this tested?

The new build code was used to generate the helper code.

## Checklist

## Related Issues
Related to: #77


hack/update-codegen.sh
```diff
@@ -0,0 +1,33 @@
...
+set -o pipefail
+
+SCRIPT_ROOT=$(cd -- "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
+CODEGEN_PKG=${1:-"${SCRIPT_ROOT}/bin"}
+THIS_PKG="github.com/llm-d/llm-d-inference-payload-processor"
+
+source "${CODEGEN_PKG}/kube_codegen.sh"
+
+kube::codegen::gen_helpers \
+    --boilerplate "/dev/null" \
```
nirrozenbaum, line 28: nit: don't we want to add boilerplate.txt to inject the header to the generated files?

nirrozenbaum:
/lgtm
/approve

left one nit about boilerplate, no need to block can be handled in follow up

hack/update-codegen.sh
shmuelk: In the llm-d-inference-scheduler it was decided that we don't need the copyright statement on generated code that we really didn't write.

nirrozenbaum: I'm fine with that :)
so no need for a follow up on this point

Issue #81, opened 2026-05-12
[Feature]: add TTFT extractor to datalayer


Mohammad-nassar10:
### Problem Statement

Add a TTFTExtractor that measures time-to-first-token per model.

### Proposed Solution

- Capture `FirstTokenTimestamp` on the first response body chunk in `HandleResponseBody`
- Add `TTFT time.Duration` to `ResponsePayload` (computed as `FirstTokenTimestamp - RequestReceivedTimestamp`)
- Write TTFT avg to model's "ttft" attribute in the `DataStore`

### Willingness to Contribute

Yes, I can submit a PR

Issue #76, opened 2026-05-12
[Feature]: Prefix based model scorer


ronenkat:
### Problem Statement

TTFT is affected by the amount of pre-fill that needs to be done.
However, requests with overlapping prefixes may be sent to different models by the filter->score->pick pipeline when there is more than one choice.
Sending a request to be run by a model that has seen a prefix would result in faster inference time.

### Proposed Solution

1. Collect prompt prefixes from request (in response processing) storing the prefixes in the datalayer
2. prompt prefix calculation will be done in the data layer extractors, total prefixes will be capped.
3. a prompt prefix scorer (running in the model selector) will produce a normalized score between [0...1], 0 reflecting no model has seen a prefix of this request, and 1 reflecting that the whole request has been seen by a model.

### Alternatives Considered

1. Querying actual kv-cache content from the model-serving pods - adds overheads and latency

### Willingness to Contribute

Yes, I can submit a PR

nirrozenbaum:
need to make sure we don't replicate kv-cache manager repo. 
maybe it does make sense to extend kv-cache for understanding what prefix exists in general in the pool? 
sounds like the easier path rather than rewriting everything, worth exploration.

ronenkat:
> need to make sure we don't replicate kv-cache manager repo. maybe it does make sense to extend kv-cache for understanding what prefix exists in general in the pool? sounds like the easier path rather than rewriting everything, worth exploration.

My starting point approach is to reuse the approximate prefix plugin in the EPP refactored to models instead of endpoints.

PR #84, opened 2026-05-12
Add datastore access from the handle to enable plugins to read and update the datastore

ronenkat:
## What does this PR do?

Enable plugins to access the datastore using the handle mechanism.

## Why is this change needed?

Plugins, like #81, need to update the model attributes in order to support calculations in Scorers.
Providing access via the Handle which is already defined in the plugin mechanism is a simple and decoupled mechanism.

## How was this tested?

## Checklist

## Related Issues

Examples of features that require access to datastore #76 #81


pkg/framework/handle.go
```diff
@@ -30,12 +32,14 @@ type Handle interface {
...
+	datastore.Datastore
 }
 
 // payloadProcessorHandle is an implementation of the Handle interface.
 type payloadProcessorHandle struct {
-	ctx context.Context
-	mgr ctrl.Manager
+	ctx       context.Context
+	mgr       ctrl.Manager
+	datastore datastore.Datastore
```
Mohammad-nassar10: `datastore` now refers to the field, not the package (any future reference to `datastore.NewStore()` or other package-level symbols inside the struct methods would silently shadow.
```suggestion
	ctx context.Context
	mgr ctrl.Manager
	ds  datastore.Datastore
```

pkg/framework/handle.go
```diff
@@ -51,9 +55,22 @@ func (h *payloadProcessorHandle) ReconcilerBuilder() *ctrlbuilder.Builder {
...
+	return h.datastore.Models()
+}
+
+func NewHandle(ctx context.Context, mgr ctrl.Manager, ds datastore.Datastore) Handle {
 	return &payloadProcessorHandle{
-		ctx: ctx,
-		mgr: mgr,
+		ctx:       ctx,
+		mgr:       mgr,
+		datastore: ds,
```
Mohammad-nassar10:
```suggestion
		ctx: ctx,
		mgr: mgr,
		ds:  ds,
```

pkg/framework/handle.go
```diff
@@ -30,12 +32,14 @@ type Handle interface {
 	Context() context.Context
 	Client() client.Client
 	ReconcilerBuilder() *ctrlbuilder.Builder
+	datastore.Datastore
```
Mohammad-nassar10: What do you think about exposing `Datastore() datastore.Datastore` as a single method on Handle instead of embedding `datastore.Datastore` directly? This keeps the Handle interface surface smaller (1 method instead of 3), makes the dependency explicit at the call `(handle.Datastore().GetOrCreateModel(...))`, and means any future methods added to the datastore interface are automatically available without needing to update `Handle`.

nirrozenbaum:
/hold

I don’t think plugins should have have access to datastore through handle. this creates spaghetti code.

ronenkat:
> /hold
> 
> I don’t think plugins should have have access to datastore through handle. this creates spaghetti code.

can you elaborate?
The current way to access the information in the datastore is through global variables.

nirrozenbaum:
> > /hold
> > I don’t think plugins should have have access to datastore through handle. this creates spaghetti code.
> 
> can you elaborate? The current way to access the information in the datastore is through global variables.

model selector plugins should assume the models are received as input to the plugin. 
more specifically:
https://github.com/llm-d/llm-d-inference-payload-processor/blob/1a39b3e837180ef610f831c629c2a8d0a27ea3bb/pkg/framework/modelselector/plugins.go#L29

https://github.com/llm-d/llm-d-inference-payload-processor/blob/1a39b3e837180ef610f831c629c2a8d0a27ea3bb/pkg/framework/modelselector/plugins.go#L38

Filter and Scorer interfaces get as input `models []datalayer.Model`. 
the plugins should assume a different component, internal to IPP, created a snapshot of the models from datastore before the call. each entry in this models slice can be used by `model.GetAttributes()`, which can be used to access any information stored in datastore by other plugins.

plugins shouldn't be allowed to work directly with datastore which is an IPP internal storage and is not meant to be read/updated/shared directly with custom plugins that could be written by external users. 

example scorer from @davidbreitgand recent PR (which is still open, but just to emphasize the idea):
https://github.com/davidbreitgand/llm-d-inference-payload-processor/blob/df35ab11bd639b19cc17026a49235bebcb2dbb38/pkg/framework/modelselector/scorer/costaware/plugin.go#L112

does it make sense?

cmd/runner/runner.go
```diff
@@ -170,7 +170,8 @@ func (r *Runner) Run(ctx context.Context) error {
 		}
 	}
 
-	handle := framework.NewHandle(ctx, mgr)
+	ds := datastore.NewStore()
```
nirrozenbaum: this should be `inmemory.NewDatastore()`

pkg/plugins/basemodelextractor/base_model_to_header_test.go
```diff
@@ -116,7 +117,8 @@ func TestBaseModelToHeaderPluginFactory(t *testing.T) {
 			}
 
 			// Create a handle using the test manager
-			handle := framework.NewHandle(context.Background(), mgr)
+			ds := datastore.NewStore()
```
nirrozenbaum: ditto

pkg/framework/datalayer/datasource/types.go
```diff
@@ -16,69 +16,7 @@ limitations under the License.
...
-	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework"
-)
-
 // DataStore provides model-keyed access to aggregated runtime metrics.
 type DataStore interface {
 	GetOrCreateModel(name string) Model
 }
-
-// DataSource is the base interface for background data layer components.
-type DataSource interface {
```
nirrozenbaum, line 32: was this removed intentionally?

ronenkat: moved to new location to avoid circular dependencies.
added

pkg/plugins/datalayer/notificationsource/plugin.go
```diff
@@ -26,7 +26,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework"
-	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/datalayer"
+	datalayer "github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/datalayer/datasource"
```
Mohammad-nassar10: This can be confusing, using `dlsrc` as other files is better.

pkg/framework/handle.go
```diff
@@ -30,12 +31,14 @@ type Handle interface {
 	Context() context.Context
 	Client() client.Client
 	ReconcilerBuilder() *ctrlbuilder.Builder
+	Datastore() datastore.Datastore
```
Mohammad-nassar10, line 34: Which option is better: `Handle.Datastore()` returns `datastore.Datastore` from pkg/datastore, or a narrower interface defined in `pkg/framework/datalayer` (as we had before)?
In the first option, `pkg/framework` now imports `pkg/datastore` (an implementation package), which inverts the usual dependency direction (maybe this will create circular dependency in the future?). It also exposes `DeleteModel` and `Models()` to all plugins, which can be hidden from them in the second option.

pkg/framework/datalayer/datasource/types.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package datalayer
+package datasource
```
nirrozenbaum, line 17: why did we move it to datasource?
this now reads datasource.Extractor, while datasource is one concept and extractor in another concept. 
before this change it was - `datalayer.DataSource` and `datalayer.Extractor`

pkg/plugins/datalayer/inflightrequests/plugin_test.go
```diff
@@ -88,16 +68,16 @@ func getInflightRequests(t testing.TB, ds *fakeDataStore, model string) Inflight
 	return rc
 }
 
-func newInflightRequestsTest(t *testing.T) (*InflightRequestsExtractor, *fakeDataStore) {
+func newInflightRequestsTest(t *testing.T) (*InflightRequestsExtractor, datastore.Datastore) {
 	t.Helper()
-	ds := newFakeDataStore()
+	ds := inmemory.NewDatastore()
```
nirrozenbaum: actually if we want to isolate tests to the plugin, we should use mock datastore (otherwise bugs in ds may leak into these tests causing failures). 
unit test should mock all dependencies to test only the unit.

pkg/framework/handle.go
ronenkat: If I read correctly, you are suggesting to move the interface from `datastore.Datastore` to `pkg/framework/datalayer`...

pkg/framework/datalayer/datasource/types.go
ronenkat: `datastore.go `imports `datalayer` (for model)
`datalayer` imports pkg/framework
`pkg/framework/handle.go` import `datastore`
Which created circular import. A solution was to move the types.go into `datalayer.DataSource`

pkg/framework/handle.go
Mohammad-nassar10: Not quite. The suggestion is to restore the separate, narrow interface that we had in `pkg/framework/datalayer..,` (with just one get method). and change the return value in `handle` to this interface (`datalayer.datastore`). WDYT?

pkg/framework/datalayer/datasource/types.go
nirrozenbaum: got it

Mohammad-nassar10: What about breaking this circular import by moving the `Model` (with the `AttributeMap`) into `pkg/framework` for example (they are more "general" concepts)? Then no need to move the rest to `datasource`.

nirrozenbaum: +1. 
we should probably follow the same structure as epp with framework/interfaces and framework/plugins to avoid the circular dependencies

pkg/framework/handle.go
ronenkat: We will have plugins that require more than just getting data for a specific model.
We would end up with the same interface duplicated.

pkg/framework/datalayer/datasource/types.go
ronenkat: @nirrozenbaum Sounds good.
I will create a follow-up issue and PR to address the refactoring, as it goes beyond updating the handle.

pkg/plugins/basemodelextractor/base_model_to_header_test.go
```diff
@@ -116,7 +117,8 @@ func TestBaseModelToHeaderPluginFactory(t *testing.T) {
 			}
 
 			// Create a handle using the test manager
-			handle := framework.NewHandle(context.Background(), mgr)
+			ds := inmemory.NewDatastore()
```
shmuelk: The mock for the Datastore should be used here as well

pkg/plugins/basemodelextractor/base_model_to_header_test.go
```diff
@@ -117,8 +116,7 @@ func TestBaseModelToHeaderPluginFactory(t *testing.T) {
 			}
 
 			// Create a handle using the test manager
-			ds := inmemory.NewDatastore()
-			handle := framework.NewHandle(context.Background(), mgr, ds)
+			handle := framework.NewHandle(context.Background(), mgr, nil)
```
shmuelk: Why pass nil and not an instance of the fakeDatastore?

pkg/plugins/datalayer/inflightrequests/plugin_test.go
```diff
@@ -89,7 +88,7 @@ func makeResponseEvent(model string, durationMs int, maxTokens float64) dlsrc.Ev
 }
 
 // getInflightRequests asserts the inflight-requests attribute exists for model and returns it.
-func getInflightRequests(t testing.TB, ds datastore.Datastore, model string) InflightRequestsCount {
+func getInflightRequests(t testing.TB, ds *fakeDataStore, model string) InflightRequestsCount {
```
shmuelk: Please revert this change.  fakeDataStore should be an instance of datastore.Datastore.

pkg/plugins/datalayer/inflightrequests/plugin_test.go
```diff
@@ -102,7 +101,7 @@ func getInflightRequests(t testing.TB, ds datastore.Datastore, model string) Inf
 	return rc
 }
 
-func newInflightRequestsTest(t *testing.T) (*InflightRequestsExtractor, datastore.Datastore) {
+func newInflightRequestsTest(t *testing.T) (*InflightRequestsExtractor, *fakeDataStore) {
```
shmuelk: Please revert this change.  fakeDataStore should be an instance of datastore.Datastore.

ronenkat:
Thanks @shmuelk. updated.

pkg/datastore/fakedatastore.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (f *fakeDataStore) Models() []string {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	names := make([]string, 0, len(f.models))
+	for name := range f.models {
+		names = append(names, name)
+	}
+	return names
+}
```
shmuelk, line 40: +1

PR #91, opened 2026-05-13
Adds generic helper to read model keys with automated type assertion

### LINKED ISSUE #88: [Feature]: add a generic type asserting attribute read to datalayer.Model
### Problem Statement

Adding it will avoid multiple recurring type assertions when reading model attributes in plugins.

### Proposed Solution

The solution will be similar to [what is proposed here](https://github.com/llm-d/llm-d-inference-payload-processor/pull/71#discussion_r3222050796).

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

davidbreitgand:
## What does this PR do?

Adds a generic template read helper that reads `model` attributes by a given key and asserts the value to the type T.
Returns an error if the key is not found or the type assertion fails.

## Why is this change needed?

Helps to avoid recurring type assertions in plugins using `model` attributes.

## How was this tested?

- [x] Unit tests added/updated

## Checklist

## Related Issues

Fixes #88

### DIFF
--- a/pkg/framework/datalayer/model.go
+++ b/pkg/framework/datalayer/model.go
@@ -16,6 +16,8 @@ limitations under the License.
 
 package datalayer
 
+import "fmt"
+
 type Model interface {
 	GetName() string
 	GetAttributes() AttributeMap
@@ -42,4 +44,22 @@ func (m *model) GetName() string {
 
 func (m *model) GetAttributes() AttributeMap {
 	return m.attributes
+}
+
+// Reads model attribute by a given key and asserts the value to the type T.
+// Returns an error if the key is not found or the type assertion fails.
+func ReadModelKey[T any](m Model, key string) (T, error) {
+	var zero T
+
+	raw, ok := m.GetAttributes().Get(key)
+	if !ok {
+		return zero, fmt.Errorf("attribute %q: not found", key)
+	}
+
+	val, ok := raw.(T)
+	if !ok {
+		return zero, fmt.Errorf("unexpected type for key %q: got %T", key, raw)
+	}
+
+	return val, nil
 }
--- a/pkg/framework/datalayer/model.go
+++ b/pkg/framework/datalayer/model.go
@@ -16,8 +16,6 @@ limitations under the License.
 
 package datalayer
 
-import "fmt"
-
 type Model interface {
 	GetName() string
 	GetAttributes() AttributeMap
@@ -44,22 +42,4 @@ func (m *model) GetName() string {
 
 func (m *model) GetAttributes() AttributeMap {
 	return m.attributes
-}
-
-// Reads model attribute by a given key and asserts the value to the type T.
-// Returns an error if the key is not found or the type assertion fails.
-func ReadModelKey[T any](m Model, key string) (T, error) {
-	var zero T
-
-	raw, ok := m.GetAttributes().Get(key)
-	if !ok {
-		return zero, fmt.Errorf("attribute %q: not found", key)
-	}
-
-	val, ok := raw.(T)
-	if !ok {
-		return zero, fmt.Errorf("unexpected type for key %q: got %T", key, raw)
-	}
-
-	return val, nil
 }
--- a/pkg/framework/datalayer/model_test.go
+++ b/pkg/framework/datalayer/model_test.go
@@ -16,10 +16,7 @@ limitations under the License.
 
 package datalayer
 
-import (
-	"strings"
-	"testing"
-)
+import "testing"
 
 // Verifies non-nil model
 // Verifies name is preserved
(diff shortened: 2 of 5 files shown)


pkg/framework/datalayer/model.go
```diff
@@ -43,3 +45,21 @@ func (m *model) GetName() string {
 func (m *model) GetAttributes() AttributeMap {
 	return m.attributes
 }
+
+// Reads model attribute by a given key and asserts the value to the type T.
+// Returns an error if the key is not found or the type assertion fails.
+func ReadModelKey[T any](m Model, key string) (T, error) {
```
ronenkat: can we use `ReadModelAttribute` to be synced with `GetAttributes`

ronenkat (commented):
Nice.

pkg/framework/datalayer/model.go
```diff
@@ -43,3 +45,21 @@ func (m *model) GetName() string {
...
+	var zero T
+
+	raw, ok := m.GetAttributes().Get(key)
```
nirrozenbaum: nice. 
one minor comment about the code (should operate on attributesmap instead of a model):

```suggestion
func ReadAttribute[T any](attributeMap AttributeMap, key string) (T, error) {
	var zero T

	raw, ok := attributeMap.Get(key)
```
also, can we move this function to attributemap file?
if it operates on attributemap this is where it belongs..

pkg/framework/datalayer/model_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"strings"
+	"testing"
+)
+
+// Verifies non-nil model
+// Verifies name is preserved
+// Verifies attributes are initialized
+func TestNewModel(t *testing.T) {
```
nirrozenbaum, line 24: and then this test become irrelevant, we're not testing model, we're testing that's it's possible to read key from attributemap with generics.

pkg/framework/datalayer/model_test.go
```diff
@@ -0,0 +1,89 @@
...
+// Verifies success case: key is found and value is of expected type.
+// Verifies error case: key is not found.
+// Verifies error case: value is not of expected type.
+func TestReadModelKey(t *testing.T) {
+	t.Run("success", func(t *testing.T) {
+		m := NewModel("test-model")
+		expected := testCloneableValue{Value: 42}
+		m.GetAttributes().Put("score", expected)
+
+		got, err := ReadModelKey[testCloneableValue](m, "score")
```
nirrozenbaum: and this test moves to the attributemap_test.go..

pkg/framework/datalayer/model.go
```diff
@@ -43,3 +45,21 @@ func (m *model) GetName() string {
...
+	if !ok {
+		return zero, fmt.Errorf("attribute %q: not found", key)
+	}
+
+	val, ok := raw.(T)
+	if !ok {
+		return zero, fmt.Errorf("unexpected type for key %q: got %T", key, raw)
```
Mohammad-nassar10: nit: adding the expected type for easier debugging.
```suggestion
		return zero, fmt.Errorf("attribute %q: expected %T, got %T", key, zero, raw)
```

nirrozenbaum:
I pushed a similar PR to llm-d-router (for attributesmap, as requested in the review). 
xref here: https://github.com/llm-d/llm-d-router/pull/1163

pkg/framework/datalayer/model.go
davidbreitgand: Good point, will move. Does it make sense to add a wrapper in the `model.go` to prevent verbosity or is it confusing?

pkg/framework/datalayer/model_test.go
davidbreitgand: @nirrozenbaum Do you want a wrapper in the `model.go` as well to limit verbosity, or do you want the assertion only to address the attributemap?

nirrozenbaum: mmmm not sure if it’s needed. 
we can start with ReadAttribute function and see if more is needed

pkg/framework/datalayer/model.go
davidbreitgand: Thanks. Fixed.

davidbreitgand:
@nirrozenbaum : fixed. 

- The generic type assertion is only in the `attributemap`. Can add a wrapper to the model to prevent verbosity, but it might be confusing to a user which interface to use. So, leaving only in the `attributemap`. 
- Enhanced to handle both values and pointers (following similar issues in [PR#1163](https://github.com/llm-d/llm-d-router/pull/1163)
- Added tests to attributemap
- Removed unneeded tests from model

pkg/framework/datalayer/model_test.go
davidbreitgand: Sure.

Issue #15, opened 2026-04-27
introduce payload processing profiles


nirrozenbaum:
as we start to use payload processing plugins, we are seeing increasing demand for having to run different plugins in different scenarios.
for example:
- when the model is specified in the request, we might need to run one set of plugins. 
- when the model is not specified and the mode is set to "auto-selection", we need to run a different set of plugins which includes in particular a plugin for model selection based on the request properties and current state of the supported models.

this requires being able to define multiple profiles, and have a `profile-resolver` plugin which, should be the first one to run, analyze the the request/state, and decide dynamically on the selected profile.

nirrozenbaum:
related to #77

nirrozenbaum:
cc: @shmuelk related to config api

szedan-rh:
@nirrozenbaum / @noyitz - will take it

nirrozenbaum:
@szedan-rh this is actually already implemented.. can be closed.

PR #92, opened 2026-05-13
design: payload processing profiles and profile picker

noyitz:
## What does this PR do?

Adds a design document for introducing profiles and a profile picker to the IPP (issue #15). This is a **design-only PR** — no code changes. The goal is to iterate on the architecture before implementation.

The design describes:
- **Three-stage pipeline**: pre-processing → profile selection → profile execution
- **Profiles**: named, ordered chains of request + response plugins
- **Profile picker**: a plugin that selects which profile to run based on the enriched request
- **Pre-processing**: shared plugins that always run before profile selection
- **Model selector integration**: sits inside profiles as a regular RequestProcessor plugin

## Why is this change needed?

Today the IPP runs the same plugin chain for every request. Different request types (model specified, auto-selection, priority routing) need different processing paths. Profiles enable this.

## How was this tested?

Design document only — no code to test.

## Checklist

## Related Issues

Relates to #15
Related config work: #77

### DIFF
--- a/designs/015-payload-processing-profiles/README.md
+++ b/designs/015-payload-processing-profiles/README.md
@@ -3,6 +3,7 @@
 Author(s): @noyitz
 
 Related issue: [#15](https://github.com/llm-d/llm-d-inference-payload-processor/issues/15)
+Related proposals: [Proposal 043 — ModelSelector Architecture](https://github.com/llm-d/llm-d-inference-payload-processor/tree/main/docs/proposals/043-model-selection-framework)
 
 ## Summary
 
@@ -120,15 +121,60 @@ If the picker returns an unknown profile name or an error, the framework returns
 
 ## Model Selector Integration
 
-The model selector is not a special concept at the profile level. It is a RequestProcessor plugin that happens to internally run its own Filter → Score → Pick pipeline. It sits in a profile's request chain alongside other request plugins.
+The ModelSelector ([Proposal 043](https://github.com/llm-d/llm-d-inference-payload-processor/tree/main/docs/proposals/043-model-selection-framework)) is a RequestProcessor plugin that internally runs its own Filter → Score → Pick pipeline to select the best model for a request. It is not a special concept at the profile level — it sits in a profile's request chain alongside other request plugins.
 
-This means:
+### How the ModelSelector works internally
 
-- Different profiles can use different model selector instances with different strategies (cost-optimized, quality-optimized, fallback)
-- The execution order between model selection and other plugins is explicit — you can see exactly where in the chain model selection happens
-- Profiles without model selection simply don't include a model selector plugin
+The ModelSelector has its own layered architecture that mirrors the IPP profile structure:
 
-The model selector can itself support multiple internal profiles with its own profile selection (same recursive pattern), but that is internal to the model selector and transparent to the IPP profile system.
+- **ModelSelector** — the top-level component. Receives candidate models and runs a ModelSelectorProfile to select the best one.
+- **ModelSelectorProfile** — an ordered pipeline of Filter → Score → Pick phases, analogous to an IPP profile's ordered plugin chain.
+- **Filter plugins** — remove models that are unfit for selection (e.g., rate-limited, unavailable). Zero or more per profile.
+- **Scorer plugins** — score remaining models in a normalized [0,1] range with configurable weights (e.g., CostScorer, LatencyScorer). Zero or more per profile.
+- **Picker plugin** — selects the winning model from scored candidates (e.g., MaxScorePicker picks the highest score, WeightedRandomPicker samples proportionally to scores). Exactly one per profile.
+
+The ModelSelector receives a CycleState from the IPP pipeline, allowing it to share data with other plugins in the same profile (e.g., reading model metadata written by a pre-processing plugin).
+
+### How it fits in the IPP profile
+
+The ModelSelector appears as a single step in a profile's request chain. From the IPP profile's perspective, it is just another RequestProcessor — the internal Filter/Score/Pick pipeline is transparent:
+
+```
+Profile: auto-routing
+  Request Chain:
+    1. model-selector (internally runs: Filter → Score → Pick → writes selected model)
+    2. model-provider-resolver (reads selected model, resolves provider)
+    3. api-translation (translates to provider format)
+    4. apikey-injection (injects provider credentials)
+```
+
+### Multiple model selectors across profiles
+
+Different profiles can use different ModelSelector instances, each configured with different Filter/Scorer/Picker plugins:
+
+- A **standard** profile might not include a ModelSelector at all — the user specified the model directly
+- An **auto-routing** profile might include a cost-optimized ModelSelector (CostScorer + MaxScorePicker)
+- A **priority** profile might include a quality-optimized ModelSelector (QualityScorer + MaxScorePicker)
+
+This is one of the key benefits of treating the ModelSelector as a regular plugin — it composes naturally with the profile system without any special-casing.
+
+### Recursive pattern
+
+The ModelSelector can itself support multiple internal profiles with its own profile selection — the same pattern as the IPP profile system, applied recursively. For example, a single ModelSelector plugin could internally select between a cost-optimized scoring profile and a quality-optimized scoring profile based on request properties. This is internal to the ModelSelector and transparent to the IPP profile that contains it.
+
+This recursive pattern — profiles containing plugins that themselves contain profiles — is a deliberate design choice. It means we design the profile/picker pattern once and reuse it at both the IPP level and the ModelSelector level.
+
+### Current implementation status
+
+The ModelSelector framework is implemented and merged:
+
+- **ModelSelector.Select()** — the entry point that runs a single profile ([PR #72](https://github.com/llm-d/llm-d-inference-payload-processor/pull/72), merged)
+- **ModelSelectorProfile.Run()** — executes the Filter → Score → Pick pipeline with per-plugin latency metrics
+- **WeightedScorer** — wraps scorers with configurable weights for score aggregation
+- **MaxScorePicker** — picks the highest-scored model with random tie-breaking ([PR #74](https://github.com/llm-d/llm-d-inference-payload-processor/pull/74))
+- **WeightedRandomPicker** — selects proportionally to scores using the A-Res sampling algorithm ([PR #74](https://github.com/llm-d/llm-d-inference-payload-processor/pull/74))
+
+Filter and scorer plugin implementations (cost, latency, rate-limit) are planned as separate issues.
 
 ## Relationship to Config API
 
--- a/designs/015-payload-processing-profiles/README.md
+++ b/designs/015-payload-processing-profiles/README.md
@@ -130,8 +130,10 @@ The ModelSelector has its own layered architecture that mirrors the IPP profile 
 - **ModelSelector** — the top-level component. Receives candidate models and runs a ModelSelectorProfile to select the best one.
 - **ModelSelectorProfile** — an ordered pipeline of Filter → Score → Pick phases, analogous to an IPP profile's ordered plugin chain.
 - **Filter plugins** — remove models that are unfit for selection (e.g., rate-limited, unavailable). Zero or more per profile.
-- **Scorer plugins** — score remaining models in a normalized [0,1] range with configurable weights (e.g., CostScorer, LatencyScorer). Zero or more per profile.
-- **Picker plugin** — selects the winning model from scored candidates (e.g., MaxScorePicker picks the highest score, WeightedRandomPicker samples proportionally to scores). Exactly one per profile.
+- **Scorer plugins** — score remaining models in a normalized [0,1] range with configurable weights (e.g., CostScorer, LatencyScorer). Zero or more per profile. When multiple scorers are configured, each scorer's output is multiplied by its weight and accumulated into a single weighted score per model.
+- **Picker plugin** — selects the winning model based on the weighted scores produced by the scorers. The picker receives the full list of scored models and uses those scores to make its decision. For example, MaxScorePicker selects the model with the highest weighted score, while WeightedRandomPicker uses the scores as probabilities — a model scored 0.9 is much more likely to be selected than one scored 0.1. Exactly one per profile.
+
+The data flows through the pipeline sequentially: filters narrow the candidate set, scorers evaluate and rank the remaining candidates, and the picker makes the final selection using those rankings.
 
 The ModelSelector receives a CycleState from the IPP pipeline, allowing it to share data with other plugins in the same profile (e.g., reading model metadata written by a pre-processing plugin).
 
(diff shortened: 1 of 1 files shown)


noyitz:
@nirrozenbaum @shmuelk design document for profiles and profile picker — would appreciate your feedback. This is runtime architecture only, intentionally doesn't touch config format (that's #77).

designs/015-payload-processing-profiles/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Proposed Architecture
+
+The IPP pipeline is organized into three stages:
+
+### Stage 1: Pre-Processing
+
+A set of shared plugins that always run for every request, regardless of which profile is selected. These plugins enrich the request with information that the profile picker needs to make its decision.
+
+Pre-processing is necessary because the profile picker often cannot decide at the start of the request. For example, the picker may need to know whether the user specified a model or requested auto-selection. But the model name lives in the request body — it needs to be extracted by `body-field-to-header` before the picker can check for it. Without pre-processing, the picker would have to duplicate body-parsing logic, breaking the plugin composability model.
```
nirrozenbaum, line 48: this is very much related to how we implement profile picker. 
IMO we should *start as any other plugin* and not CEL. there are use cases where it would be very hard to implement profile picker with limited language some as regex/CEL/other. 

for example, if one wants to select an external model ONLY when all internal models are saturated. 
the profile picker should be able to take into account the request properties **as well as the system state**. 

given that profile picker is like any other plugin - it should have access to all fields like any other plugin, so pre-processing plugins are not required for that purpse. 

pre-processing plugins might make sense anyway if we want to define that certain plugins should always run, no matter which profile was selected. but the reason for that is not to enrich data for profile picker, but rather just a set of common plugins that represent common behavior for all use cases.

designs/015-payload-processing-profiles/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+- The request (after pre-processing plugins have run)
+- CycleState (shared state from pre-processing)
+- The set of available profile names
+
+It returns the name of the profile to execute.
+
+The profile picker runs exactly once per request. It does not iterate or re-pick (unlike the upstream scheduler's ProfileHandler which can iteratively select profiles based on previous results). If iterative profile selection is needed in the future, the interface can be extended.
+
+The profile picker's decision logic should be adjustable at runtime without restarting the IPP. For example, an operator should be able to change which requests route to which profiles (e.g., shifting traffic from a cost-optimized profile to a quality-optimized profile) by updating configuration. This requires the picker to support reloadable decision logic — whether through hot-reloadable config, CEL expressions, or another mechanism that doesn't require recompilation or restart.
```
nirrozenbaum, line 119: I doubt it would be possible to express all possible configurations in those mechanisms.
maybe a better path is to support HA if we use redis as distributed datastore, which would allow rolling restarts while not loosing state.

designs/015-payload-processing-profiles/README.md
```diff
@@ -0,0 +1,169 @@
...
+- The response chain of the selected profile handles any post-processing
+
+The scheduler's iterative pattern could be added later if a use case emerges, but for the initial implementation a single pick is sufficient.
+
+## Open Questions
+
+1. **Should response plugins also be profile-specific, or always shared?** The current design makes them profile-specific. An alternative is to have shared post-processing (like shared pre-processing) that always runs regardless of profile. This could be useful for plugins that should run for every response — metrics collection, security validation, audit logging. A shared post-processing stage would mirror pre-processing and ensure these concerns are never accidentally omitted from a profile.
+
+2. **Shared pre/post-processing symmetry** — If we support shared pre-processing on the request side, should we also support shared post-processing on the response side? The use cases are clear: metrics collection, security plugins, audit logging all need to run for every response regardless of which profile handled the request.
+
```
nirrozenbaum, line 166: aren't these points the same?

nirrozenbaum (commented):
finished reviewing the doc. 
it feels more like a motivation & very high level concepts. 

the main point is that there are no clear action items derived from the doc.

designs/015-payload-processing-profiles/README.md
```diff
@@ -0,0 +1,169 @@
...
+Today the IPP runs a single, fixed chain of plugins for every request:
+
+```
+Request → [RequestPlugin₁ → RequestPlugin₂ → ... → RequestPluginₙ] → Dispatch → [ResponsePlugin₁ → ResponsePlugin₂ → ... → ResponsePluginₘ] → Response
+```
+
+All requests go through the same plugins in the same order. The order is determined at startup via command-line flags. There is no way to conditionally run different plugins for different requests.
+
+This works for simple use cases (extract model name, resolve provider, inject credentials). But as the IPP takes on intelligent routing responsibilities — model selection, cost-based routing, fallback — different requests need fundamentally different processing paths.
```
ronenkat, line 23: What does `resolve provider` means?

designs/015-payload-processing-profiles/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Motivation
+
+Consider these scenarios:
+
+**Scenario 1: Model specified in request**
```
ronenkat, line 29: the text does not align with the IPP as of today. i.e., resolve the provider, translate the API format and inject credentials...

designs/015-payload-processing-profiles/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+## What is a Profile Picker
+
+The profile picker is a plugin that implements a dedicated interface. It receives:
+
+- The request (after pre-processing plugins have run)
+- CycleState (shared state from pre-processing)
+- The set of available profile names
+
+It returns the name of the profile to execute.
```
ronenkat, line 115: we can run all Pre-Processing Plugins, then if a profile exist in the cycle state, run the profile.

designs/015-payload-processing-profiles/README.md
ronenkat, line 119: reading the above it start to sound like another type of filter-score-pick mechanism....

designs/015-payload-processing-profiles/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+Each scenario requires a different set of plugins in a different order. Profiles make this possible.
+
+## Proposed Architecture
+
+The IPP pipeline is organized into three stages:
+
+### Stage 1: Pre-Processing
+
+A set of shared plugins that always run for every request, regardless of which profile is selected. These plugins enrich the request with information that the profile picker needs to make its decision.
```
ronenkat, line 46:
```suggestion
A set of shared plugins that always run for every request in the order specified, regardless of which profile is selected. These plugins enrich the request with information that the profile picker needs to make its decision.
```

designs/015-payload-processing-profiles/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+Pre-processing is necessary because the profile picker often cannot decide at the start of the request. For example, the picker may need to know whether the user specified a model or requested auto-selection. But the model name lives in the request body — it needs to be extracted by `body-field-to-header` before the picker can check for it. Without pre-processing, the picker would have to duplicate body-parsing logic, breaking the plugin composability model.
+
+### Stage 2: Profile Selection
+
+A profile picker plugin examines the enriched request and selects which profile to run. The picker is a plugin that implements a defined interface — it receives the request (after pre-processing has run) and the set of available profiles, and returns the name of the selected profile.
```
ronenkat, line 52: can we make the picker just one more request processing plugin? the set of plugin can come from the configuration or the data-layer.
if no profile-selection plugin is in the pre-processing, then a default profile (i.e., the first configured will run)
if one if configured, it write the decision to cycle-state...

designs/015-payload-processing-profiles/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Relationship to Config API
+
+This design describes the runtime architecture — what components exist, how they connect, and in what order they execute. How these components are configured (YAML format, plugin references, parameter passing) is the responsibility of [#77](https://github.com/llm-d/llm-d-inference-payload-processor/issues/77).
+
+This design should be implementable regardless of the config format chosen. The key requirements for the config are:
+
+- Ability to define a set of pre-processing plugins with their order
+- Ability to define multiple named profiles, each with an ordered request chain and response chain
+- Ability to specify which plugin serves as the profile picker
```
ronenkat, line 143:
```suggestion
- A profile picker plugin must be specified in the pre-processing plugins list in order to use a profile beyond the default one.
```

designs/015-payload-processing-profiles/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+- It picks once (no iterative re-picking)
+- There is no result aggregation — only one profile runs per request
+- The response chain of the selected profile handles any post-processing
+
+The scheduler's iterative pattern could be added later if a use case emerges, but for the initial implementation a single pick is sufficient.
+
+## Open Questions
+
+1. **Should response plugins also be profile-specific, or always shared?** The current design makes them profile-specific. An alternative is to have shared post-processing (like shared pre-processing) that always runs regardless of profile. This could be useful for plugins that should run for every response — metrics collection, security validation, audit logging. A shared post-processing stage would mirror pre-processing and ensure these concerns are never accidentally omitted from a profile.
```
ronenkat, line 163: we should start simple...

designs/015-payload-processing-profiles/README.md
nirrozenbaum: generally speaking, I actually think that representing every behavior as “just another request plugin” goes the opposite direction (same for model selector). 

the framework should suggest strong types, along with validation and defaults. 

plugin should represent customized logic to hook into extension points. 
framework types should represent the built in mechanism and how to hook the plugins. 

in that context - profile picker should be a framework strong type, also model selector.

ronenkat: adopting this viewpoint, the IPP flow should be, for example:
1. a pre-processing processor (of request processing plugins)
2. a PRE-MODEL-SELECTOR profile-specific request processor, starting with a profile selector followed by running profile specific plugins (of request processing plugins)
3. a model selector processor, starting with a profile selector, selecting selector profile, running a filter-score-pick profile
4. a POST-MODEL-SELECTOR profile-specific request processor, starting with a profile selector followed by running profile specific plugins (of request processing plugins)
In this approach, the pre-processor, model-selector and profile-processor are 1st class citizens that run plugins.

designs/015-payload-processing-profiles/README.md
shmuelk: I assume figure out who is providing the model inference service....

designs/015-payload-processing-profiles/README.md
```diff
@@ -0,0 +1,169 @@
...
+Each scenario requires a different set of plugins in a different order. Profiles make this possible.
+
+## Proposed Architecture
+
+The IPP pipeline is organized into three stages:
+
+### Stage 1: Pre-Processing
+
+A set of shared plugins that always run for every request, regardless of which profile is selected. These plugins enrich the request with information that the profile picker needs to make its decision.
+
```
shmuelk, line 47: The pre-processing plugins can change the request that will eventually be sent out?

designs/015-payload-processing-profiles/README.md
shmuelk: We need to separate how the IPP framework consumes the various types of plugins vs how they are described/configured/mentioned in the configuration.

Actually I think the IPP flow is different. It is as follows:

1. pre-processing plugins run
2. The profile picker picks a profile
3. The request plugins of the profile are run. This includes the model selector, which is just another request plugin.

We need to extend things so that any plugin can reference other plugins and run them. The assistance from the framework is simply to order the instantiation of the plugins, preventing loops, etc.

designs/015-payload-processing-profiles/README.md
shmuelk: I disagree. The profile picker reference is a 1st class citizen in the configuration. If left out and there is only one profile in the configuration, a default `single-profile-picker` plugin is chosen, which simply always chooses the single profile in the configuration.

designs/015-payload-processing-profiles/README.md
nirrozenbaum: I think the intention was yes. e.g., add header with the model (header change).

nirrozenbaum: assuming this is a plugin that is always needed. 
otherwise the alternative is to specify that plugin multiple times, in all profiles.

designs/015-payload-processing-profiles/README.md
```diff
@@ -0,0 +1,169 @@
...
+**Scenario 1: Model specified in request**
+The user sends `{"model": "llama-70b"}`. The IPP needs to resolve the provider, translate the API format, and inject credentials. No model selection is needed.
+
+**Scenario 2: Auto-selection mode**
+The user sends `{"model": "auto"}`. The IPP needs to run a model selector (Filter → Score → Pick) to choose the best model, then proceed with provider resolution and credential injection.
+
+**Scenario 3: Priority routing**
+A high-priority request (indicated by a header) should use a different model selection strategy than a standard request — perhaps a quality-optimized selector instead of a cost-optimized one.
+
```
davidbreitgand, line 37: Just for completeness:
```suggestion

***Scenario 4: Batch routing**
A latency-insensitive request (indicated by a header) should use cost-optimized selector instead of the latency-optimized one.
```

designs/015-payload-processing-profiles/README.md
```diff
@@ -0,0 +1,169 @@
...
+Consider these scenarios:
+
+**Scenario 1: Model specified in request**
+The user sends `{"model": "llama-70b"}`. The IPP needs to resolve the provider, translate the API format, and inject credentials. No model selection is needed.
+
+**Scenario 2: Auto-selection mode**
+The user sends `{"model": "auto"}`. The IPP needs to run a model selector (Filter → Score → Pick) to choose the best model, then proceed with provider resolution and credential injection.
+
+**Scenario 3: Priority routing**
+A high-priority request (indicated by a header) should use a different model selection strategy than a standard request — perhaps a quality-optimized selector instead of a cost-optimized one.
```
davidbreitgand, line 36:
```suggestion
A high-priority request (indicated by a header) should use a different model selection strategy than a standard request — perhaps a latency-optimized selector instead of a cost-optimized one.
```

designs/015-payload-processing-profiles/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+The model selector is not a special concept at the profile level. It is a RequestProcessor plugin that happens to internally run its own Filter → Score → Pick pipeline. It sits in a profile's request chain alongside other request plugins.
+
+This means:
+
+- Different profiles can use different model selector instances with different strategies (cost-optimized, quality-optimized, fallback)
+- The execution order between model selection and other plugins is explicit — you can see exactly where in the chain model selection happens
+- Profiles without model selection simply don't include a model selector plugin
+
+The model selector can itself support multiple internal profiles with its own profile selection (same recursive pattern), but that is internal to the model selector and transparent to the IPP profile system.
```
davidbreitgand, line 133: @noyitz , just for my understanding: does this imply that the internal model selector profile can override the initial profile picker, and potentially these profiles will push in different directions? What are scenarios that benefit from the recursive pattern?

nirrozenbaum:
profile picker already implemented (interface + first implementation). 
closing this PR for now, if needed it can be reopened.

Issue #93, opened 2026-05-14
[Feature]: running-requests scorer


Mohammad-nassar10:
### Problem Statement

Add a simple running-requests scorer to score models based on their current in-flight request count, ranking the least loaded model highest by assigning a normalized score in [0.0, 1.0]. The most loaded model receives a score of 0.0, and the least loaded receives a score of 1.0.

### Proposed Solution

- `RunningRequestsScorer` implements the Scorer interface embedding Plugin.
- The scorer reads the `running-requests` attribute from each model, populated by `RunningRequestsExtractor` in the `datalayer`.
Models without a `running-requests` attribute are treated as idle (0 requests).
If all models have the same count (including all zero), all receive a score of 1.0.
If there is only one model, it receives a score of 1.0.

### Willingness to Contribute

Yes, I can submit a PR

PR #94, opened 2026-05-14
Adds an inflight-requests scorer

Mohammad-nassar10:
## What does this PR do?
Add a simple running-requests scorer to score models based on their current in-flight request count, ranking the least loaded model highest by assigning a normalized score in [0.0, 1.0]. The most loaded model receives a score of 0.0, and the least loaded receives a score of 1.0.

> **Note:** This PR is stacked on #75. Because both PRs target `main`,
> merge #75 first.
## How was this tested?

## Checklist

## Related Issues
Fixes issue #93.


pkg/framework/modelselector/scorer/runningrequests/plugin.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Score returns a score in [0,1] for each model based on its running request count.
+// Formula: score = (max - count) / (max - min)
+func (s *RunningRequestsScorer) Score(_ context.Context, _ *framework.CycleState, _ *framework.InferenceRequest, models []datalayer.Model) map[datalayer.Model]float64 {
+	var minCount int64 = math.MaxInt64
+	var maxCount int64 = math.MinInt64
+
+	counts := make(map[datalayer.Model]int64, len(models))
+	for _, m := range models {
+		n := runningRequests(m)
```
nirrozenbaum: nit: can you please use descriptive names? 
now the code is still small so m/n can be followed, but as soon as the codebase is getting bigger it becomes unreadable. 

for example use var names like: model, count/modelRunningRequest, etc.

pkg/framework/plugins/modelselector/scorer/inflightrequests/plugin.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+		if count > maxCount {
+			maxCount = count
+		}
+	}
+
+	scores := make(map[datalayer.Model]float64, len(models))
+	for _, model := range models {
+		if maxCount == minCount {
+			scores[model] = 1.0
```
ronenkat, line 86: What does maxCount == minCount  mean? should it mean that the scorer input is not important? and if so, then it should be 0, not 1.

pkg/framework/plugins/modelselector/scorer/inflightrequests/plugin.go
```diff
@@ -0,0 +1,105 @@
...
+		} else {
+			scores[model] = float64(maxCount-requestCounts[model]) / float64(maxCount-minCount)
```
ronenkat, line 88: Consider the following example:
Model A = 100
Model B = 1
Scores = A=0, B=1
lets be careful about the edge scenarios.

ronenkat (commented):
Some comment on calculation.
The scorer will evolve, so lets make sure the current score make sense, and we can have a running system.
We can fine-tune it further using benchmarks and evaluations.

pkg/framework/modelselector/scorer/inflightrequests/plugin.go
```diff
@@ -0,0 +1,105 @@
...
+
+// compile-time interface assertion
+var _ modelselector.Scorer = &InflightRequestsScorer{}
+
+// InflightRequestsScorer scores models based on their in-flight request count.
+// The least loaded model scores 1.0; the most loaded scores 0.0.
+// Models without an inflight-requests attribute are treated as idle (0 requests).
+// If all models have the same count, all score 1.0.
+type InflightRequestsScorer struct {
+	name framework.TypedName
```
nirrozenbaum: be consistent with plugins. this is typedName (has a type and a name).

```suggestion
	typedName framework.TypedName
```

pkg/framework/plugins/modelselector/scorer/inflightrequests/plugin.go
nirrozenbaum: as long as all models get the same score in this case it has no meaning what is the score, since it's linear formula.
e.g., if `score(A) > score(B)` then: `score(A) + xxx > score(B) + xxx`, where `xxx` is the score given by this scorer.

pkg/framework/plugins/modelselector/scorer/inflightrequests/plugin.go
nirrozenbaum: @ronenkat in your example the resulting score might make sense. 
actually the edge case here is when there are two models and they are very close in the num of requests:
Model A = 100
Model B = 99. 
result scores: A=0, B=1. 

in the extreme case both models have very close number of requests and yet, the scorer gives them scores that are not representing it properly.

pkg/framework/plugins/modelselector/scorer/inflightrequests/plugin.go
Mohammad-nassar10: The score is not important as long as they have the same score. Also, I think 1 is more "explainable" (Returning 0 when all models are equal would signal "no model is acceptable", but 1 means "all are equally preferred, pick any").

pkg/framework/plugins/modelselector/scorer/inflightrequests/plugin.go
nirrozenbaum: this is kinda similar discussion to what we had in the cost scorer. 
we could represent the score by the number of requests handled by `model-x` relative to the total number of inflight requests. 

e.g., score(model) = 1 - ( inflight(model) / total_inflight_requests ).
this will produce the following results:
Model A = 100
Model B = 1
Scores: A=`1 - (100/101) = ~0.001` , B=`1 - (1/100) = ~0.99`

Model A = 100
Model B = 99.
Scores: A=`1 - (100/199) = ~0.4975` , B=`1 - (99/199) = ~0.5025`

shmuelk:
Please add the code to register the plugin

Mohammad-nassar10:
@shmuelk Added, thanks!

pkg/framework/plugins/modelselector/scorer/inflightrequests/plugin.go
Mohammad-nassar10: We use min-max scoring here. For load balancing, the direction matters more than the magnitude, and we want to prefer the less-loaded model regardless of how close the counts are (proportional gap less matters here).

PR #106, opened 2026-05-17
[DESIGN] Prefix scoring

ronenkat:
## What does this PR do?

Proposes a Prefix Hash-Index-Score system to enable routing requests to models with cached prefixes. The system uses three plugins: PrefixHashing computes prefix hashes and queries an indexer; RequestPrefixScorer assigns higher scores to models with longer prefix matches; PrefixIndexingExtractor updates the indexer with prefix-to-model mappings.

## Why is this change needed?

Leveraging prefix caching in modern LLM inference engines can significantly reduce TTFT. This design enables intelligent routing to models that have previously processed similar request prefixes, maximizing cache hit rates.

## Checklist

- [ ] Design document is complete and addresses the problem
- [ ] Architecture and components are well-defined
- [ ] Interface contracts are specified
- [ ] Open questions are documented

## Related Issues

Describe a design for #76


docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,159 @@
+# Prefix Hash-Index-Score System
+
+## Summary
+
+The Prefix Hash-Index-Score system enables intelligent request routing to models that have previously processed similar request prefixes, thereby improving Time-To-First-Token (TTFT) by leveraging prefix caching capabilities in modern LLM inference engines. The system operates in three phases within the Inference Payload Processor (IPP): a pre-processing plugin computes prefix hashes and queries an indexer to identify models with matching cached prefixes; a scoring plugin assigns higher scores to models with longer prefix matches; and a response extraction plugin updates the indexer with information about which models have processed which prefixes.
```
davidbreitgand: Nit: typo.

```suggestion
The Prefix Hash-Index-Score system enables intelligent request routing to models that have previously processed similar request prefixes, thereby improving Time-To-First-Token (TTFT) by leveraging prefix caching capabilities in modern LLM inference engines. The system operates in three phases within the Inference Payload Processor (IPP): a pre-processing plugin computes prefix hashes and queries an indexer to identify models with matching cached prefixes; a scoring plugin assigns higher scores to models with longer prefix matched; and a response extraction plugin updates the indexer with information about which models have processed which prefixes.
```

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Proposed Architecture
+
+The Prefix Hash-Index-Score system consists of three main components that operate at different stages of the request processing pipeline within the IPP:
+
+1. **PrefixHashing Plugin** (Pre-processing, RequestProcessor)
+   - **Type**: [`framework.RequestProcessor`](pkg/framework/plugin.go)
+   - **Role**: Executes during the pre-processing phase before model selection
+   - **Responsibilities**:
+     - Computes prefix hashes for incoming requests by dividing the prompt into fixed-size blocks
```
davidbreitgand: Can we also factor in TTL? In external API providers, the default is `5` minutes before eviction. The default can also be explicitly changed. But as a first step, can we count `5` minutes from the first request with this prefix we sent to the model, and reset the count each time we route a request with that prefix to the model? Having this knowledge will help with better scoring. 

```suggestion
- Computes prefix hashes for incoming requests by dividing the prompt into fixed-size blocks
- Tracks expected TTL of the prefix.
```

docs/designs/001-prefix-scoring/README.md
aradhalevy: I suggest emphasizing that the goal is to add a Scorer. The system will allow that

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,159 @@
...
+     - Queries the prefix indexer to identify models that have previously processed matching prefixes
+     - Stores the computed hashes and model match information in the [`CycleState`](pkg/framework/cycle_state.go) for use by the scorer
+     - Implements a greedy longest-prefix matching strategy to find the best cache matches
```
aradhalevy, line 25: This sounds like a capability that logically should be under the scorer as it more than simple pre-processing and more about the scoring.

Also if this runs before the Filter it would make unnecessary computations of irrelevant models. 

I think you address this in open question 2. My gut feeling is that this is more suited in the Scorer logic but its up to you

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+2. **RequestPrefixScorer** (Scoring, Scorer)
+   - **Type**: [`modelselector.Scorer`](pkg/framework/modelselector/scorer.go)
+   - **Role**: Executes during the model selection scoring phase
+   - **Responsibilities**:
+     - Retrieves prefix hash state from [`CycleState`](pkg/framework/cycle_state.go) (stored by PrefixHashing plugin)
+     - Scores each candidate model based on its prefix cache match ratio
+     - Calculates score as: `matchedBlocks / totalBlocks` (range: 0.0 to 1.0)
```
aradhalevy, line 33: What are the TotalBlocks? the total blocks of the request? 

If it is, why is this an important parameter when calculating the "prefix match"? 

For example,  a request with with 5 MatchedBlocks and 8 TotalBlocks would get a much higher score than a request with 100 MatchedBlocks and 500 TotalBlocks. I am not sure this is what we would want.

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,159 @@
...
+     - Processes [`ResponseEventType`](pkg/framework/event.go) events to track completed requests
+     - Computes prefix hashes from the original request
+     - Updates the prefix indexer with information about which model processed which prefix hashes
+     - Maintains the indexer's knowledge of prefix-to-model mappings over time
+
+### Request Body Hashing and Indexing
+
+The hash function implementation uses the xxHash algorithm for fast, high-quality hashing.
+
+1. **Block-based Hashing**: The request prompt is divided into fixed-size text blocks. Each block is hashed independently, creating a sequence of [`BlockHash`](pkg/plugins/prefixhashing/types.go) values.
```
aradhalevy: The link in [`BlockHash`] doesn't work, is it intentionally here for when you have the implementation?

docs/designs/001-prefix-scoring/README.md
davidbreitgand, line 33: Coming back to my comment on TTL.
```suggestion
- Calculates score as: `(matchedBlocks / totalBlocks)*(TTL/maxTTL)` (range: 0.0 to 1.0)
```
Rationale: for a cache to be useful for routing, _both_  conditions must be true:

- The prefix must be cached (measured by the matched blocks to total blocks ratio)
- The cache must still be valid (measured by normalized TTL)

The TTL parameter for a specific prefix for a specific model should be [re-]started when a decision is made to route to this model. 

Since this might not ultimately be the model with the longest prefix and the longest TTL, because another scorer and a different model might be preferred, it would be beneficial to update the indexer _right after_ the routing decision is made. I think this immediate update will be equally beneficial for the proposed LRU mechanism. 

The TTL issue can be described in the design, but implementation-wise, the first step can be done without it to simplify the implementation. 

Does this make sense?

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,159 @@
...
+- Deterministic: same input always produces same hash
+- Collision-resistant: different inputs produce different hashes with high probability
+- Order-preserving: prefix relationship is maintained in hash sequences
+
+### Indexing Mechanism
+
+The indexer provides an efficient bidirectional mapping between prefix hashes and models:
+
+1. **Hash-to-Models Mapping** (`hashToModels map[BlockHash]modelSet`):
+   - Maps each [`BlockHash`](pkg/plugins/prefixhashing/types.go:36) to a set of models that have that hash cached
```
aradhalevy: Isn't each `BlockHash` unique per model as you add `model_name` for the first block's hash, and every block's hash after that is dependent on the previous one (meaning it dependes on the model name)?

If this is the case, I am not sure how the lookup of requests will happen, do you add each model name to to each request and lookup like that? 

Maybe I don't understand correctly

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,159 @@
...
+   - Enables fast lookup: "Which models have this prefix hash cached?"
+
+2. **Model-to-LRU Mapping** (`modelToLRU map[ModelID]*lru.Cache[BlockHash, struct{}]`):
```
davidbreitgand, line 79: TTL tracking vs LRU mapping.

LRU behavior depends on the cache pressure. TTL is absolute. With external API vendors, we have TTL explicitly available. With internal models, we don't have TTL, and LRU is a good approximation. Maybe we can combine the two techniques based on the visibility of metrics?

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,159 @@
...
+   - Each model has its own LRU cache tracking which hashes it has processed
+   - Default capacity: 256,144 entries per model (configurable)
+   - Automatically evicts oldest hashes when capacity is reached
+   - Approximates the actual prefix cache state on model inference servers
+
+**Eviction and Cleanup**: The indexer uses per-model LRU caches with custom eviction callbacks that automatically remove models from the hash-to-models mapping when hashes are evicted. A background cleanup goroutine runs every 2 minutes to remove entries for inactive models, preventing memory leaks. All operations are thread-safe using `sync.RWMutex` with read locks for queries and write locks for updates.
+
+**Capacity Sizing**: The initial LRU version uses a default LRU capacity (e.g., 131,072 entries) to be revised for in future versions.
+
+### 4.3 Datastore Updates
```
aradhalevy: nit: what is 4.3? by accident?

docs/designs/001-prefix-scoring/README.md
davidbreitgand: > a response extraction plugin updates the indexer with information about which models have processed which prefixes.

I am a bit concerned that at a high request rate, this might be too slow a mechanism. What if we request `RequestProcessor` pipeline that integrates `Filter -> Scorer -> Picker` to immediately update the indexer upon making the final routing decision?

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,159 @@
...
+}
+```
+
+This would provide explicit dependency injection, easier testing, and eliminate global state coupling. However, it requires framework changes that affect all plugin factories.
+
+### 2. Cycle State for Hash and Model Information
+
+**Question**: Should the computed hashes and model match information be stored in cycle state? If not, how should the pre-processing plugin pass this information to the scorer?
+
+**Suggested Implementation**: The PrefixHashing plugin stores a [`RequestHashingState`](pkg/plugins/prefixhashing/types.go:55) in the [`CycleState`](pkg/framework/cycle_state.go) containing:
```
davidbreitgand: I think something went wrong with the permalinks. I get:

Image

docs/designs/001-prefix-scoring/README.md
ronenkat: I don't see how the score should be affected by TTL.
Either it is in the prefix-cache or not.

docs/designs/001-prefix-scoring/README.md
ronenkat: Thanks. will remove the links.

docs/designs/001-prefix-scoring/README.md
ronenkat: Thanks. good catch.

docs/designs/001-prefix-scoring/README.md
ronenkat: Agree that it is a concern that we should investigate.
The current approach optimizes for repeated prompts (one agent), not concurrent ones.
we should experiment and evolve in future versions.

docs/designs/001-prefix-scoring/README.md
ronenkat: i don't think it is the prefix hashing responsibility.
Taking a note to add into the indexer management TTL aspects.

docs/designs/001-prefix-scoring/README.md
ronenkat: good point. 
My thought was that hashes may be useful for more than the prefix scorer.
but it is possible to simplify the architecture to two componets instead of three

docs/designs/001-prefix-scoring/README.md
ronenkat: yes - of the request.
we are not comparing requests.

docs/designs/001-prefix-scoring/README.md
davidbreitgand: > Also if this runs before the Filter it would make unnecessary computations of irrelevant models.

+1

docs/designs/001-prefix-scoring/README.md
ronenkat: adding TTL

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,159 @@
...
+```
+
+The [`Datastore`](pkg/datastore/datastore.go:32) interface is extended:
+
+```go
+type Datastore interface {
+    GetOrCreateModel(name string) datalayer.Model
+    DeleteModel(name string)
+    Models() []string
+    GetPrefixIndexer() PrefixIndexer  // New method
```
davidbreitgand, line 113: `GetPrefixIndexer() PrefixIndexer  // New method` tightly couples the datastore (generic) to the prefix-specific functinoality. Every `datastore` implementation must now have prefix indexing, even if not used. 

What if we separate indexing into a "service"?
```go
type PrefixIndexerService struct {
    indexer PrefixIndexer
    datastore datastore.Datastore
}
```

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,153 @@
...
+   - **Responsibilities**:
+     - Processes `ResponseEventType` events to track completed requests
+     - Updates the prefix indexer with information about which model processed which prefix hashes
+     - Maintains the indexer's knowledge of prefix-to-model mappings over time, leveraging LRU and TTL eviction policies.
+
+### Request Body Hashing and Indexing
+
+The hash function implementation uses the xxHash algorithm for fast, high-quality hashing.
+
+1. **Block-based Hashing**: The request prompt is divided into fixed-size text blocks. Each block is hashed independently, creating a sequence of `BlockHash` values.
```
davidbreitgand, line 49: `BlockHash` undefined:
```suggestion
**Block-based Hashing**: The request prompt is divided into fixed-size text blocks. Each block is hashed independently, creating a sequence of `BlockHash` values.
`type BlockHash uint64`
```

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+**Question**: Should the computed hashes and model match information be stored in cycle state? If not, how should the pre-processing plugin pass this information to the scorer?
+
+**Suggested Implementation**: The PrefixHashing plugin stores a `RequestHashingState` in the `CycleState` containing:
+- `PrefixHashes []BlockHash`: The computed prefix hashes
+- `PrefixCacheModels map[ModelID]int`: Models and their longest prefix match length
+
+The RequestPrefixScorer retrieves this state from CycleState to calculate scores.
+
+**Considerations**:
```
davidbreitgand: I think something went wrong with the permalinks. I get:

Image

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+The prefix indexer is integrated into the datastore to provide centralized access across all IPP components.
+
+**Interface Extension:**
+
+The `PrefixIndexer` interface is added to avoid import cycles:
+
+```go
+type PrefixIndexer interface {
+    // Methods defined by implementation
```
davidbreitgand: Just for my understanding. Is it going to be something like?
```go
type PrefixIndexer interface {
    Query(hashes []BlockHash) map[ModelID]int
    Update(string ModelID, hashes []BlockHash) error
}
```

Maybe consider to explicitly define in the design document?

docs/designs/001-prefix-scoring/README.md
davidbreitgand: Suppose we routed to an external model. The prefix is now in the prefix cache of the provider, and will stay there for `5` mins by default. I think we want to reward routing to this model as opposed to the one that already has the cache expired for this prefix.

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+The `Datastore` interface is extended:
+
+```go
+type Datastore interface {
+    GetOrCreateModel(name string) datalayer.Model
+    DeleteModel(name string)
+    Models() []string
+    GetPrefixIndexer() PrefixIndexer  // New method
+	 SetPrefixIndexer(indexer PrefixIndexer) PrefixIndexer // New method
```
Mohammad-nassar10, line 114: `Set` doesn't suppose to return a value. Maybe you mean to return `error`.

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,153 @@
...
+     - Scores each candidate model based on its prefix cache match ratio
+     - Calculates score as: `matchedBlocks / totalBlocks` (range: 0.0 to 1.0)
+     - Models with longer prefix matches receive higher scores, influencing routing decisions
+
+3. **PrefixIndexingExtractor** (Post-processing, Extractor)
+   - **Type**: `framework.Extractor`
+   - **Role**: Executes during the response extraction phase after request completion, running out of the response hot-path as a data layer event processing component
+   - **Responsibilities**:
+     - Processes `ResponseEventType` events to track completed requests
+     - Updates the prefix indexer with information about which model processed which prefix hashes
```
Mohammad-nassar10, line 41: Does the extractor get the hashes from `CycleState`? If yes, is `CycleState` still alive when the extractor runs post-response?
Let's specify what data is carried in `ResponseEventType` and how the extractor gets the hashes. One option is to pass the computed hashes within the `ResponsePayload`. Or, if recomputing is fast enough, it can be done in the extractor (the payload contains the body).

docs/designs/001-prefix-scoring/README.md
ronenkat: it returns the set Indexer as it support a singleton approach, run only once.

docs/designs/001-prefix-scoring/README.md
ronenkat: added

docs/designs/001-prefix-scoring/README.md
ronenkat: Agree that our current datastore implementation is for pre-defined attributes.
Please explain your suggestion, and how PrefixIndexerService will help share indexing as a global state?

docs/designs/001-prefix-scoring/README.md
ronenkat: @Mohammad-nassar10 does it make sense to update the `ResponseEventType` to pass the `RequestContext` ?

Mohammad-nassar10: Yes, this makes sense. (I understood that you mean to add the `RequestContext` with the `CycleState` to the `ResposnePayload`).

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,160 @@
...
+   ```
+
+2. **Chained Hashing**: To ensure that identical prefixes in different positions produce different hashes, each block hash incorporates the previous block's hash:
+   ```
+   hash(block[i]) = xxhash(block[i].content + hash(block[i-1]))
+   ```
+   This creates a dependency chain where changing any block affects all subsequent hashes.
+
+3. **Configurable Parameters**:
+   - `HashBlockSize`: Size of each text block (default: 64)
```
kfirtoledo, line 62: Is depend on the vLLM GPU block size it can be 16,64, maybe you can pull it as a metric from the inference scheduler or vLLM

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+The indexer provides an efficient bidirectional mapping between prefix hashes and models:
+
+1. **Hash-to-Models Mapping** (`hashToModels map[BlockHash]modelSet`):
+   - Maps each `BlockHash` to a set of models that have that hash cached
+   - Enables fast lookup: "Which models have this prefix hash cached?"
+
+2. **Model-to-LRU Mapping** (`modelToLRU map[ModelID]*lru.Cache[BlockHash, struct{}]`):
+   - Each model has its own LRU and TTL cache tracking which hashes it has processed
+   - Default capacity: 256,144 entries per model (configurable)
```
kfirtoledo, line 81: Same here, I think you can also take it as a metric from the inference scheduler

docs/designs/001-prefix-scoring/README.md
```diff
@@ -0,0 +1,160 @@
...
+   - Automatically evicts oldest hashes when capacity is reached
+   - Approximates the actual prefix cache state on model inference servers
```
kfirtoledo, line 83: Are you also going to consider CPU offloading?

docs/designs/001-prefix-scoring/README.md
ronenkat: The idea is to have an approximation of the cache.
The notion of blocks is to walk the decision tree faster.

docs/designs/001-prefix-scoring/README.md
ronenkat: Right, it would make sense to understand how much KVs are kept.

docs/designs/001-prefix-scoring/README.md
ronenkat: that could be an optimization to consider different KV locations with different weight.

Issue #114, opened 2026-05-19
perf: ModelSelectorProfile allocates memory per-model on every request


yehuditkerido:
### Problem Statement

`ModelSelectorProfile.Run` is called on every inference request. Profiling shows
that allocations scale linearly with the number of candidate models:

- 5 models → 39 allocations
- 25 models → 67 allocations  
- 100 models → 142 allocations

Two potential causes identified:

1. **Logging arguments boxed even when disabled** — `logger.V(DEBUG).Info(...)`
   evaluates and boxes all arguments before checking if the log level is enabled

2. **Per-model heap allocations** — each `&ScoredModel{}` in the scoring loop
   creates a separate allocation instead of using a contiguous backing array

At scale (many models, high request rate), this creates unnecessary GC pressure
and latency.

nirrozenbaum:
can we extend the point (1) analysis and detect logging arguments across the whole repo?

nirrozenbaum:
closing as done

PR #118, opened 2026-05-20
perf(modelselector): reduce per-request memory allocations in hot path

yehuditkerido:
## What does this PR do?

Reduces memory allocations in the model selection hot path by:
1. Guarding verbose/debug logs with `Enabled()` checks to avoid argument boxing
2. Using a contiguous backing array for `ScoredModel` instead of per-model heap allocations

## Why is this change needed?

Per-request allocations scaled with model count, causing GC pressure. Now ~82% fewer allocations (142 → 26 for 100 models) and constant regardless of model count. 

## How was this tested?

Added `BenchmarkModelSelectorProfileRun` to track allocation regression. All tests pass.

## Related Issues

#114

### DIFF
--- a/pkg/handlers/request.go
+++ b/pkg/handlers/request.go
@@ -110,14 +110,23 @@ func (s *Server) HandleRequestBody(ctx context.Context, reqCtx *RequestContext, 
 
 // runRequestPlugins executes request plugins in the order they were registered.
 func (s *Server) runRequestPlugins(ctx context.Context, cycleState *plugin.CycleState, request *requesthandling.InferenceRequest) error {
+	logger := log.FromContext(ctx).V(logutil.DEFAULT)
+
+	// Cache verbose logger and check Enabled() once to avoid per-iteration
+	// allocations from argument boxing when logging at that level is disabled.
+	verboseLogger := logger.V(logutil.VERBOSE)
+	verboseEnabled := verboseLogger.Enabled()
+
 	var err error
 	for _, plugin := range s.requestPlugins {
-		log.FromContext(ctx).V(logutil.VERBOSE).Info("Executing request plugin", "plugin", plugin.TypedName())
+		if verboseEnabled {
+			verboseLogger.Info("Executing request plugin", "plugin", plugin.TypedName())
+		}
 		before := time.Now()
 		err = plugin.ProcessRequest(ctx, cycleState, request)
 		metrics.RecordPluginProcessingLatency(requestPluginExtensionPoint, plugin.TypedName().Type, plugin.TypedName().Name, time.Since(before))
 		if err != nil {
-			log.FromContext(ctx).V(logutil.DEFAULT).Error(err, "Failed to execute request plugin", "plugin", plugin.TypedName())
+			logger.Error(err, "Failed to execute request plugin", "plugin", plugin.TypedName())
 			return err
 		}
 	}
--- a/pkg/handlers/response.go
+++ b/pkg/handlers/response.go
@@ -132,14 +132,23 @@ func (s *Server) HandleResponseTrailers(trailers *eppb.HttpTrailers) ([]*eppb.Pr
 
 // runResponsePlugins executes response plugins in the order they were registered.
 func (s *Server) runResponsePlugins(ctx context.Context, cycleState *plugin.CycleState, response *requesthandling.InferenceResponse) error {
+	logger := log.FromContext(ctx).V(logutil.DEFAULT)
+
+	// Cache verbose logger and check Enabled() once to avoid per-iteration
+	// allocations from argument boxing when logging at that level is disabled.
+	verboseLogger := logger.V(logutil.VERBOSE)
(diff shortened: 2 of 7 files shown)


pkg/handlers/request.go
```diff
@@ -109,14 +109,23 @@ func (s *Server) HandleRequestBody(ctx context.Context, reqCtx *RequestContext,
 
 // runRequestPlugins executes request plugins in the order they were registered.
 func (s *Server) runRequestPlugins(ctx context.Context, cycleState *framework.CycleState, request *framework.InferenceRequest) error {
+	logger := log.FromContext(ctx)
```
nirrozenbaum: set the verbosity level once at the beginning and reuse the logger
```suggestion
	logger := log.FromContext(ctx).V(logutil.DEFAULT)
```

pkg/handlers/request.go
```diff
@@ -109,14 +109,23 @@ func (s *Server) HandleRequestBody(ctx context.Context, reqCtx *RequestContext,
...
+
+	// Cache logger and check Enabled() once to avoid per-iteration allocations
+	// from argument boxing when logging at that level is disabled.
+	verbose := logger.V(logutil.VERBOSE)
```
nirrozenbaum: verbose logger?

```suggestion
	verboseLogger := logger.V(logutil.VERBOSE)
```

pkg/handlers/response.go
```diff
@@ -131,14 +131,23 @@ func (s *Server) HandleResponseTrailers(trailers *eppb.HttpTrailers) ([]*eppb.Pr
 
 // runResponsePlugins executes response plugins in the order they were registered.
 func (s *Server) runResponsePlugins(ctx context.Context, cycleState *framework.CycleState, response *framework.InferenceResponse) error {
+	logger := log.FromContext(ctx)
```
nirrozenbaum: same comments

pkg/modelselector/model_selector_profile.go
```diff
@@ -155,34 +155,64 @@ func (p *ModelSelectorProfile) Run(ctx context.Context, request *framework.Infer
...
+		storage[i] = modelselector.ScoredModel{Model: model, Score: 0}
+		scoredModels[model.GetName()] = &storage[i]
 	}
 
+	// Cache loggers and check Enabled() once to avoid per-iteration allocations
+	// from argument boxing when logging at that level is disabled.
+	verbose := logger.V(logutil.VERBOSE)
+	verboseEnabled := verbose.Enabled()
+	debug := logger.V(logutil.DEBUG)
+	debugEnabled := debug.Enabled()
```
nirrozenbaum: as a rule of thumb, it would probably make sense to always put these variables at the beginning of each function

pkg/modelselector/model_selector_profile_bench_test.go
```diff
@@ -0,0 +1,131 @@
...
+// ModelSelectorProfile.Run hot path (filter → score → pick) at different
+// model counts. This benchmark serves as a regression guard for per-request
+// allocations in the model selection framework.
+//
+// Run:
+//
+//	go test -run='^$' -bench=BenchmarkModelSelectorProfileRun -benchmem -count=5 \
+//	    ./pkg/modelselector/ | tee bench.out
+//	benchstat bench.out
+func BenchmarkModelSelectorProfileRun(b *testing.B) {
```
nirrozenbaum, line 47: is the expectation to run this manually before/after PR and compare?

nirrozenbaum (commented):
some updates to model selector were done.. moved to a different package and added also picker plugins, can you please rebase and add the same also for picker?

sorry about that! 🙏

pkg/modelselector/model_selector_profile_bench_test.go
yehuditkerido: Will submit a follow up PR with CI job for that.

pkg/modelselector/model_selector_profile_bench_test.go
```diff
@@ -0,0 +1,139 @@
...
+	scores := make(map[datalayer.Model]float64, len(models))
+	for i, m := range models {
+		// Produce varied but deterministic scores
+		scores[m] = float64(i%10) / 10.0
+	}
+	return scores
+}
+
+// benchPicker is a minimal picker that selects the highest-scored model.
+type benchPicker struct {
```
nirrozenbaum, line 122: can we reuse max-score-picker instead of re-implementing max-score?
or is there any benefit from using this picker?

yehuditkerido: The minimal picker is intentional - we want to measure framework overhead, not picker logic. Using the real picker would make it harder to distinguish framework regressions from picker regressions. That said, we could add a separate benchmark for the real pickers if needed. WDYT?

nirrozenbaum: let's keep it as is for now

nirrozenbaum:
/lgtm
/approve

Thanks!

PR #121, opened 2026-05-20
Second step in re-organizing the code base

shmuelk:
## What does this PR do?
The remaining plugins and plugin related interfaces have been moved under ppkg/framework/plugins and pkg/framework/interfaces, respectively.

## Why is this change needed?
Second step in re-organizing the code base to make it easier to find things and to avoid import loops.

## How was this tested?

## Checklist


pkg/datastore/datastore.go
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 
 package datastore
 
-import "github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/datalayer"
+import "github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/datalayer"
```
ronenkat: @shmuelk can we move the datastore.go into the datalayer?
it would make sense to have it co-hosted with the rest of the data layer artifacts.

ronenkat (approved):

PR #127, opened 2026-05-24
Updates to Feature/model selector plugin 

ronenkat:
**What type of PR is this?**
/kind feature

Add model-selector as plugin (continued work on 97)


pkg/framework/plugins/requesthandling/modelselector/plugin.go
```diff
@@ -0,0 +1,138 @@
...
+func NewModelSelectorPlugin(ds datalayer.Datastore) (*ModelSelectorPlugin, error) {
+	if ds == nil {
+		return nil, fmt.Errorf("datastore is required for '%s' plugin", ModelSelectorPluginType)
+	}
+
+	// Build profile with picker only.
+	// Filters and scorers can be added via WithFilters() / WithScorers()
+	// once implementations are available.
+	profile := ms.NewModelSelectorProfile().
+		WithPicker(&defaultPicker{})
```
shmuelk: From a conceptual perspective this is confusing.

I thought there were no model selector profiles. The are simply IPPprofiles in which there can be several configurations of various plugins including the model selector plugin.

pkg/framework/plugins/requesthandling/modelselector/plugin_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/requesthandling"
+)
+
+// fakeHandle implements plugin.Handle for unit tests, providing only a Datastore.
+type fakeHandle struct{ ds datalayer.Datastore }
+
+func (f *fakeHandle) Context() context.Context                { return context.Background() }
+func (f *fakeHandle) Client() client.Client                   { return nil }
+func (f *fakeHandle) ReconcilerBuilder() *ctrlbuilder.Builder { return nil }
+func (f *fakeHandle) Datastore() datalayer.Datastore          { return f.ds }
```
shmuelk, line 47: As was done with the FakeDataStore, this should be  refactored into a common FakeHandle...

ronenkat: Lets wait with this refactoring.
I can see that some test code will need to configure plugins, which will require creating multiple different instances for the test handle implementation.

pkg/framework/plugins/requesthandling/modelselector/plugin.go
ronenkat: @shmuelk the profile here is actually a pipeline of plugins.
I suggest we rename `model_selector_profile` to be `model_selector_pipeline` to avoid the confusion.
Different `IPPprofiles` may include different  filters/scorers/pickers.
as `IPPprofiles` will be available, we will need to revise ModelSelectorPluginFactory to configure the pipeline.

ronenkat: Updated the code to use the plugins from the IPP profile handle to feed the profile.
The plugins are stored as model-selector profile. It follows the profile design in IPP

pkg/framework/plugins/requesthandling/modelselector/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+
+// NewModelSelectorPlugin creates a ModelSelector RequestProcessor plugin.
+// Candidate models are read from the Datastore on each request.
+// Filter, Scorer, and Picker plugins are sourced from the handle; if no Picker is present,
+// MaxScorePicker is used as the default.
+func NewModelSelectorPlugin(handle plugin.Handle) (*ModelSelectorPlugin, error) {
+	if handle.Datastore() == nil {
+		return nil, fmt.Errorf("datastore is required for '%s' plugin", ModelSelectorPluginType)
+	}
+
```
shmuelk: This should not happen and does not to be checked

pkg/framework/plugins/requesthandling/modelselector/plugin_test.go
```diff
@@ -0,0 +1,292 @@
...
+	f.plugins[name] = p
+}
+func (f *fakeHandle) GetAllPlugins() []fwkplugin.Plugin {
+	result := make([]fwkplugin.Plugin, 0, len(f.plugins))
+	for _, p := range f.plugins {
+		result = append(result, p)
+	}
+	return result
+}
+func (f *fakeHandle) GetAllPluginsWithNames() map[string]fwkplugin.Plugin { return f.plugins }
```
shmuelk, line 60: This will return nil, if called before AddPlugin above...

pkg/framework/plugins/requesthandling/modelselector/plugin_test.go
```diff
@@ -0,0 +1,292 @@
...
+func (f *fakeHandle) Context() context.Context                { return context.Background() }
+func (f *fakeHandle) Client() client.Client                   { return nil }
+func (f *fakeHandle) ReconcilerBuilder() *ctrlbuilder.Builder { return nil }
+func (f *fakeHandle) Datastore() datalayer.Datastore          { return f.ds }
+
+func (f *fakeHandle) Plugin(name string) fwkplugin.Plugin { return f.plugins[name] }
+func (f *fakeHandle) AddPlugin(name string, p fwkplugin.Plugin) {
+	if f.plugins == nil {
+		f.plugins = map[string]fwkplugin.Plugin{}
+	}
```
shmuelk: Instead of this if statement, add a newFakeHandle function that creates fakeHandle correctly

pkg/framework/plugins/requesthandling/modelselector/plugin_test.go
```diff
@@ -0,0 +1,292 @@
...
+	for _, name := range modelNames {
+		ds.GetOrCreateModel(name)
+	}
+	return ds
+}
+
+// TestProcessRequestWritesSelectedModelToBodyAndCycleState checks that the selected model is written to both the request body field "model" and CycleState.
+func TestProcessRequestWritesSelectedModelToBodyAndCycleState(t *testing.T) {
+	ds := newTestDatastore("model-a", "model-b", "model-c")
+	plugin, err := ModelSelectorPluginFactory(ModelSelectorPluginType, json.RawMessage(`{}`), &fakeHandle{ds: ds})
```
shmuelk: See comment above about newwFakeHandle()


```suggestion
	plugin, err := ModelSelectorPluginFactory(ModelSelectorPluginType, json.RawMessage(`{}`), NewFakeHandle("model-a", "model-b", "model-c")
```

pkg/framework/plugins/requesthandling/modelselector/plugin_test.go
```diff
@@ -0,0 +1,292 @@
...
+	ds := datastore.NewFakeDataStore()
+	for _, name := range modelNames {
+		ds.GetOrCreateModel(name)
+	}
+	return ds
+}
+
+// TestProcessRequestWritesSelectedModelToBodyAndCycleState checks that the selected model is written to both the request body field "model" and CycleState.
+func TestProcessRequestWritesSelectedModelToBodyAndCycleState(t *testing.T) {
+	ds := newTestDatastore("model-a", "model-b", "model-c")
```
shmuelk: Remove and embed in newFakeHandle function

pkg/framework/plugins/requesthandling/modelselector/plugin_test.go
```diff
@@ -0,0 +1,292 @@
...
+	if storedModel != selectedModel {
+		t.Errorf("CycleState model %q does not match body model %q", storedModel, selectedModel)
+	}
+}
+
+// TestProcessRequestSelectsFromDatastoreModels checks that the selected model is one of the candidates registered in the datastore.
+func TestProcessRequestSelectsFromDatastoreModels(t *testing.T) {
+	candidates := []string{"llama-70b", "llama-8b", "mistral-7b"}
+	ds := newTestDatastore(candidates...)
+	plugin, err := ModelSelectorPluginFactory(ModelSelectorPluginType, json.RawMessage(`{}`), &fakeHandle{ds: ds})
```
shmuelk:
```suggestion
	candidates := []string{"llama-70b", "llama-8b", "mistral-7b"}
	handle := newFAkeHandle(candidates...)
	plugin, err := ModelSelectorPluginFactory(ModelSelectorPluginType, json.RawMessage(`{}`), handle)
```

pkg/framework/plugins/requesthandling/modelselector/plugin_test.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+	if !found {
+		t.Errorf("selected model %q is not in datastore models %v", selectedModel, candidates)
+	}
+}
+
+// TestProcessRequestFailsWithEmptyDatastore checks that ProcessRequest returns an error when no candidate models are available.
+func TestProcessRequestFailsWithEmptyDatastore(t *testing.T) {
+	ds := newTestDatastore() // no models
+	plugin, err := ModelSelectorPluginFactory(ModelSelectorPluginType, json.RawMessage(`{}`), &fakeHandle{ds: ds})
```
shmuelk:
```suggestion
	plugin, err := ModelSelectorPluginFactory(ModelSelectorPluginType, json.RawMessage(`{}`), newFakeHandle())
```

pkg/framework/plugins/requesthandling/modelselector/plugin_test.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+// TestModelSelectorPluginFactoryRejectsNilDatastore checks that the factory returns an error when the handle has no datastore.
+func TestModelSelectorPluginFactoryRejectsNilDatastore(t *testing.T) {
+	_, err := ModelSelectorPluginFactory(ModelSelectorPluginType, json.RawMessage(`{}`), &fakeHandle{ds: nil})
+	if err == nil {
+		t.Fatal("expected error for nil datastore")
+	}
+}
+
```
shmuelk: Remove

pkg/framework/plugins/requesthandling/modelselector/plugin_test.go
```diff
@@ -0,0 +1,292 @@
...
+		t.Error("factory did not wire the datastore from the handle")
+	}
+}
+
+// TestTypedName checks that the plugin's TypedName type matches the registered ModelSelectorPluginType constant.
+func TestTypedName(t *testing.T) {
+	ds := newTestDatastore("model-a")
+	plugin, _ := ModelSelectorPluginFactory(ModelSelectorPluginType, json.RawMessage(`{}`), &fakeHandle{ds: ds})
+	if plugin.TypedName().Type != ModelSelectorPluginType {
+		t.Errorf("expected type %q, got %q", ModelSelectorPluginType, plugin.TypedName().Type)
```
shmuelk: It is generally a bad idea to redefine import name.

```suggestion
	handle := newFakeHandle("model-a")
	thePlugin, _ := ModelSelectorPluginFactory(ModelSelectorPluginType, json.RawMessage(`{}`), handle)
	if thePlugin.TypedName().Type != ModelSelectorPluginType {
		t.Errorf("expected type %q, got %q", ModelSelectorPluginType, thePlugin.TypedName().Type)
```

pkg/framework/plugins/requesthandling/modelselector/plugin_test.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+}
+
+// TestBuildProfileUsesDefaultMaxScorePickerWhenNoPickerInHandle checks that MaxScorePicker is used as the default picker when no picker plugin is in the handle.
+func TestBuildProfileUsesDefaultMaxScorePickerWhenNoPickerInHandle(t *testing.T) {
+	handle := &fakeHandle{ds: newTestDatastore("model-a")}
```
shmuelk:
```suggestion
	handle := &NeFakeHandle("model-a")}
```

pkg/framework/plugins/requesthandling/modelselector/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+	return &ModelSelectorPlugin{
+		typedName: plugin.TypedName{Type: ModelSelectorPluginType, Name: ModelSelectorPluginType},
+		selector:  ms.NewModelSelector(profile),
+		datastore: handle.Datastore(),
+	}, nil
+}
+
+// buildModelSelectorProfile inspects all plugins in the handle and adds them to the profile
```
shmuelk: I thought there weren't any ModelSelector profiles?

ronenkat: The profile follows the IPP profile.
Any filter, scorer and picker configured in IPP profile configuration is configured into the model-selector.
There is no NEW profile configuration.

pkg/framework/plugins/requesthandling/modelselector/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+// ModelSelectorPluginFactory is the factory function for the ModelSelector RequestProcessor plugin.
+func ModelSelectorPluginFactory(name string, _ json.RawMessage, handle plugin.Handle) (plugin.Plugin, error) {
+	return NewModelSelectorPlugin(handle)
+}
+
+// NewModelSelectorPlugin creates a ModelSelector RequestProcessor plugin.
+// Candidate models are read from the Datastore on each request.
+// Filter, Scorer, and Picker plugins are sourced from the handle; if no Picker is present,
+// MaxScorePicker is used as the default.
+func NewModelSelectorPlugin(handle plugin.Handle) (*ModelSelectorPlugin, error) {
```
nirrozenbaum: let's keep handle only in factory and this function should get:
datastore + ModelSelectorProfile. 
building the profile can be done in the factory function. 

This should make our lives much easier in tests (unit/integration), where we don't need to create handle, but only call the `NewModelSelectorPlugin`

pkg/framework/plugins/requesthandling/modelselector/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	logutil "github.com/llm-d/llm-d-inference-payload-processor/pkg/common/observability/logging"
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/datalayer"
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/modelselector"
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/plugin"
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/requesthandling"
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/plugins/modelselector/picker/maxscore"
+	ms "github.com/llm-d/llm-d-inference-payload-processor/pkg/modelselector"
```
nirrozenbaum, line 31: nit: can we remove the `ms` alias and use `modelselector`?

pkg/framework/plugins/requesthandling/modelselector/plugin.go
nirrozenbaum: +1

pkg/framework/plugins/requesthandling/modelselector/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+
+// buildModelSelectorProfile inspects all plugins in the handle and adds them to the profile
+// via AddPlugins. Scorers are pre-wrapped with a default weight of 1.0 as required by AddPlugins.
+// If no Picker plugin is found, MaxScorePicker is used as the default.
+func buildModelSelectorProfile(handle plugin.Handle) (*ms.ModelSelectorProfile, error) {
+	profile := ms.NewModelSelectorProfile()
+
+	var hasPicker bool
+	var pluginsToAdd []plugin.Plugin
+	for _, plug := range handle.GetAllPlugins() {
```
nirrozenbaum: this is wrong. 
if we decide to instantiate multiple model selector instances, each with different sub-plugins (filter/scorer/picker) this will work wrong. 

the model selector should get the ref to its plugins in the parameters:
```yaml
  - type: model-selector
     parameters:
        plugins:
         - pluginRef: rate-limit-filter
         - pluginRef: cost-scorer
            weight: 1
         - pluginRef: max-score-picker
```

pkg/framework/plugins/requesthandling/modelselector/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+type ModelSelectorPlugin struct {
+	typedName plugin.TypedName
+	selector  *ms.ModelSelector
+	datastore datalayer.Datastore
+}
+
+func (p *ModelSelectorPlugin) TypedName() plugin.TypedName {
+	return p.typedName
+}
+
```
nirrozenbaum, line 69: you're missing the `WithName` function

pkg/framework/plugins/requesthandling/modelselector/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+	result, err := p.selector.Select(ctx, request, cycleState, candidateModels)
+	if err != nil {
+		return fmt.Errorf("model selection failed: %w", err)
+	}
+
+	selectedName := result.TargetModel.GetName()
+	logger.V(logutil.VERBOSE).Info("Model selected", "model", selectedName)
+
+	cycleState.Write(SelectedModelKey, selectedName)
```
nirrozenbaum: why do we need this in cycle state?
all plugins have access to the body (in memory) and can read the model

pkg/framework/plugins/requesthandling/modelselector/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+	return nil
+}
+
+// loadCandidateModels reads all known models from the Datastore.
+func (p *ModelSelectorPlugin) loadCandidateModels() []datalayer.Model {
+	modelNames := p.datastore.Models()
+	candidates := make([]datalayer.Model, len(modelNames))
+	for i, name := range modelNames {
+		candidates[i] = p.datastore.GetOrCreateModel(name)
+	}
```
nirrozenbaum, line 115: this emphasize the fact that datastore interface is wrongly defined. 
we shouldn't need to loop over all models and get them one by one on every request. 
instead the datastore interface should have `GetModels(predicate)` call (and when we want all models, predicate == true).

pkg/framework/plugins/requesthandling/modelselector/plugin.go
nirrozenbaum: I'll change the naming from ModelSelectorProfile to ModelSelectorConfig, as @ronenkat's comment about ambiguity is correct,  using `profile` can be confusing here.
can do that in a followup not to interrupt this PR

pkg/framework/plugins/requesthandling/modelselector/plugin.go
ronenkat: For multiple model-selectors we would use profiles.
The filters, scorer and picker will be in the IPP profile, and would be automatically be picked into the model-selector w/o any additional config.
At this time, IPP support only one profile, so the list of coming directly from the plugin list in the handle, that will need to change when multiple profiles are supported.

pkg/framework/plugins/requesthandling/modelselector/plugin.go
ronenkat: Capturing as an issue

pkg/framework/plugins/requesthandling/modelselector/plugin.go
ronenkat: we already use `modelselector `to reference `pkg/modelselector`

pkg/framework/plugins/requesthandling/modelselector/plugin.go
ronenkat: Sounds good.

pkg/framework/plugins/requesthandling/modelselector/plugin.go
nirrozenbaum: right. I see now that modelselector is referenced both from interface and from plugins. 
seems like something is wrong with the abstractions here. 
it's all a result of the `buildModelSelector` function which IMO shouldn't be here. 
the plugin should reference only `"github.com/llm-d/llm-d-inference-payload-processor/pkg/modelselector"`.

the `buildModelSelector` function is kinda duplicating ModelSelectorProfile `AddPlugins`. 

the plugin should be changed to include in its paramters the list of modelselector plugin (filter/scorer/picker) and then call ModelSelectorProfile.AddPlugin(...plugins) instead of reimplementing it.

pkg/framework/plugins/requesthandling/modelselector/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+
+var _ requesthandling.RequestProcessor = &ModelSelectorPlugin{}
+
+// ModelSelectorPluginFactory is the factory function for the ModelSelector RequestProcessor plugin.
+func ModelSelectorPluginFactory(name string, _ json.RawMessage, handle plugin.Handle) (plugin.Plugin, error) {
+	profile, err := buildModelSelectorProfile(handle)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build model selector profile: %w", err)
+	}
+	return NewModelSelectorPlugin(profile, handle.Datastore()), nil
```
Mohammad-nassar10:
```suggestion
	return NewModelSelectorPlugin(profile, handle.Datastore()).WithName(name), nil
```

pkg/framework/plugins/requesthandling/modelselector/plugin.go
shmuelk: Why add tec hnical debt, if this can be done the "right way" up front

pkg/framework/plugins/requesthandling/modelselector/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/requesthandling"
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/plugins/modelselector/picker/maxscore"
+	ms "github.com/llm-d/llm-d-inference-payload-processor/pkg/modelselector"
+)
+
+const (
+	ModelSelectorPluginType = "model-selector"
+
+	// CycleState key where the selected model name is stored for downstream plugins.
+	SelectedModelKey = "selected-model"
```
Mohammad-nassar10: Is this still used?

pkg/framework/plugins/requesthandling/modelselector/plugin.go
ronenkat: sounds good. Although it would not change the imports

pkg/framework/plugins/requesthandling/modelselector/plugin.go
```diff
@@ -0,0 +1,154 @@
...
+	if err := profile.AddPlugins(pluginsToAdd...); err != nil {
+		return nil, err
+	}
+
+	if !hasPicker {
+		profile.WithPicker(maxscore.NewMaxScorePicker())
+	}
+
+	return profile, nil
+}
```
nirrozenbaum: this section is duplicating code from other parts and is mixing responsibilities. 
default values -> in the scope of config loader not here. (e.g., default picker).
creating WeightedScorer if weight appears in the config -> in the scope of config loader as well.
checking which interface a plugin implements -> in the scope of ModelSelector AddPlugins function, and should not be coded here.

this should be written differently to completely remove the reference to `interface/modelselector`

a simple possible version:
```suggestion
func buildModelSelectorPipeline(handle plugin.Handle, cfg *ModelSelectorPluginConfig) (*ms.ModelSelectorProfile, error) {
	profile := ms.NewModelSelectorProfile()

	for _, ref := range cfg.Plugins {
		plugin := handle.Plugin(ref.PluginRef)
		if plugin == nil {
			return nil, fmt.Errorf("plugin '%s' not found", ref.PluginRef)
		}
		
		if err := profile.AddPlugins(plugin); err != nil {
		    return nil, err
	    }
	}

	return profile, nil
}
```

cc also: @shmuelk

pkg/framework/plugins/requesthandling/modelselector/config.go
```diff
@@ -0,0 +1,45 @@
...
+// Weight is required when the referenced plugin is a Scorer.
+type PluginRefConfig struct {
+	PluginRef string   `json:"pluginRef"`
+	Weight    *float64 `json:"weight,omitempty"`
+}
+
+// ModelSelectorPluginConfig holds the configuration for the ModelSelectorPlugin.
+type ModelSelectorPluginConfig struct {
+	Plugins []PluginRefConfig `json:"plugins,omitempty"`
+}
```
nirrozenbaum: something here is not modeled properly. it looks like config loader code leaks into the plugin cause it's a "plugin of plugins". 
this should be handled properly at the config loading level

pkg/modelselector/model_selector_profile.go
```diff
@@ -57,6 +57,21 @@ type ModelSelectorProfile struct {
...
+// Scorers returns the weighted scorer plugins registered in the profile.
+func (p *ModelSelectorProfile) Scorers() []*WeightedScorer {
+	return p.scorers
+}
+
+// Picker returns the picker plugin registered in the profile.
+func (p *ModelSelectorProfile) Picker() modelselector.Picker {
+	return p.picker
+}
+
```
nirrozenbaum, line 74: why is this needed?

pkg/modelselector/modelselector.go
```diff
@@ -43,6 +43,11 @@ type ModelSelector struct {
 	profile *ModelSelectorProfile
 }
 
+// Profile returns the ModelSelectorProfile used by this selector.
+func (s *ModelSelector) Profile() *ModelSelectorProfile {
+	return s.profile
+}
```
nirrozenbaum, line 49: why is this needed?

pkg/framework/plugins/requesthandling/modelselector/plugin.go
ronenkat: this comment contradicts your previous review, which was asking to configure the model-selector plugins via the parameters section.
```
  - type: model-selector
     parameters:
        plugins:
         - pluginRef: rate-limit-filter
         - pluginRef: cost-scorer
            weight: 1
         - pluginRef: max-score-picker
```

ronenkat: I will move the model-selector plugin configuration back to the profile.

nirrozenbaum: this is still "configuring the model selector plguins via parameters section".
it's just a question of "which component is responsible for reading the config and create from it the right type".
in this case I think we're probably missing some functionality because most of the config loader code was taken as is from epp, whlie here we need to ability to configure plugin of plugins.

I think we should handle that in the config loader part and not in every plugin that would act as plugin of plugins.

pkg/modelselector/modelselector.go
ronenkat: profile is private and needs to be accessed from the Model-Selector plugin.
The issue is that we keep the code in two separate locations.

pkg/modelselector/model_selector_profile.go
ronenkat: The configuration uses it to set a default Picker if not set

pkg/framework/plugins/requesthandling/modelselector/config.go
shmuelk: There are some things that can be done, to make things like this more generic.

I have created an Issue on the llm-d-router with the start of some code to work on this there.

PR #128, opened 2026-05-24
add polling-source data layer plugin

ronenkat:
**What type of PR is this?**

/kind feature

This PR adds the [`pollingsource`](pkg/framework/plugins/datalayer/pollingsource/plugin.go) data layer plugin, modeled after the existing notification source pattern, and implements the [`dlsrc.PollingSource`](pkg/framework/plugins/datalayer/pollingsource/plugin.go:37) interface for polling-based collection. It allows multiple collectors to be registered with independent polling frequencies via [`PollingSource.RegisterCollector()`](pkg/framework/plugins/datalayer/pollingsource/plugin.go:77), then starts one polling loop per collector in [`PollingSource.Start()`](pkg/framework/plugins/datalayer/pollingsource/plugin.go:84). This is needed to support polling-driven model discovery where different collectors must refresh at different intervals.

Fixes #


shmuelk:
I would it would be better to have a more generic collector that can be connected with all sorts of data sources.

Such a Collector should be built in to the IPP and used automatically when polling data sources are referenced.

ronenkat:
@shmuelk updated.

pkg/framework/plugins/datalayer/pollingsource/plugin.go
```diff
@@ -0,0 +1,128 @@
...
+	defer p.mu.Unlock()
+	p.collectors = append(p.collectors, collectorEntry{collector: c, frequency: frequency})
+}
+
+// Start launches one polling goroutine per registered Collector. Returns an error if called more than once.
+func (p *PollingSource) Start(ctx context.Context) error {
+	if !p.started.CompareAndSwap(false, true) {
+		return errors.New("PollingSource already started")
+	}
+	ctx, p.cancel = context.WithCancel(ctx)
```
aradhalevy, line 152: The `p.cancel` is not under a mutex.
If you send `Start` and right after `Stop`, `Stop` could see
`if p.cancel != nil {` (cancel is nil)

and right after that	`ctx, p.cancel = context.WithCancel(ctx)` is executed and cancel is set.

pkg/framework/plugins/datalayer/pollingsource/plugin.go
```diff
@@ -0,0 +1,128 @@
...
+		p.cancel()
+		p.wg.Wait()
+	}
+}
+
+func (p *PollingSource) runCollector(ctx context.Context, c dlsrc.Collector, freq time.Duration) {
+	defer p.wg.Done()
+	logger := log.FromContext(ctx).WithName("polling-source")
+
+	ticker := time.NewTicker(freq)
```
aradhalevy, line 210: This line might panic if `freq <= 0`, probably best to add a guard

pkg/framework/plugins/datalayer/pollingsource/plugin.go
```diff
@@ -0,0 +1,128 @@
...
+	return &PollingSource{
+		name: plugin.TypedName{Type: PluginType, Name: name},
+	}, nil
+}
+
+func (p *PollingSource) TypedName() plugin.TypedName { return p.name }
+
+// RegisterCollector adds a Collector to be polled at the given frequency.
+// Safe to call before Start.
+func (p *PollingSource) RegisterCollector(c dlsrc.Collector, frequency time.Duration) {
```
aradhalevy, line 125: This is safe to call before start, but not possible to call after `Start`. If you call this after you call `Start`, the new collector won't be registered in the snapshot that `Start` saw. 
If this is intentional maybe we can add an error when trying to register after `Start` was already called.

pkg/framework/plugins/datalayer/pollingsource/plugin.go
```diff
@@ -0,0 +1,128 @@
...
+	if p.cancel != nil {
+		p.cancel()
+		p.wg.Wait()
+	}
+}
+
+func (p *PollingSource) runCollector(ctx context.Context, c dlsrc.Collector, freq time.Duration) {
+	defer p.wg.Done()
+	logger := log.FromContext(ctx).WithName("polling-source")
+
```
Mohammad-nassar10, line 191: The first poll won't fire until after `freq` time. Should we add a poll immediately on the start before entering the tick loop?
```suggestion
// Poll immediately so data is available before the first tick.
if _, err := c.Poll(ctx); err != nil {
    logger.Error(err, "initial poll failed", "collector", c.TypedName())
}
```

pkg/framework/plugins/datalayer/pollingsource/plugin_test.go
```diff
@@ -0,0 +1,100 @@
...
+	c := &fakeCollector{name: "c1"}
+	src.RegisterCollector(c, 10*time.Millisecond)
+
+	if err := src.Start(context.Background()); err != nil {
+		t.Fatalf("Start: %v", err)
+	}
+	time.Sleep(50 * time.Millisecond)
+	src.Stop()
+
+	if got := c.pollCount.Load(); got < 2 {
```
Mohammad-nassar10: `time.Sleep(50ms)` + fixed count threshold can fail falsely on loaded CI, the ticker goroutine may not get scheduled often enough during that window. A more reliable pattern waits until the condition is met or a hard timeout expires.

pkg/framework/plugins/datalayer/pollingsource/plugin.go
ronenkat: Thanks. Revised.

pkg/framework/plugins/datalayer/pollingsource/plugin.go
ronenkat: added support for registering after, although no way to use it at the moment,

pkg/framework/plugins/datalayer/pollingsource/plugin_test.go
ronenkat: removing the test

pkg/framework/plugins/datalayer/pollingsource/plugin.go
ronenkat: updated

Mohammad-nassar10 (approved):

PR #129, opened 2026-05-24
Add model list configuration collector

ronenkat:
**What type of PR is this?**

/kind feature

This PR introduces a new polling collector plugin `modelconfigcollector` that reads model configurations from a JSON file and registers them in the datastore. The collector:

- Implements the `Collector` interface from the data layer framework
- Reads a JSON configuration file containing a list of model names
- Periodically polls the configuration file and updates the datastore with model entries using the `GetOrCreateModel` interface
- Registers the plugin in the runner's in-tree plugin registry

This enables dynamic model configuration management through file-based configuration, allowing models to be added or updated without code changes.

Fixes #


pkg/framework/plugins/datalayer/modelconfigcollector/plugin.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, err
+	}
+
+	var cfg ModelsConfig
+	if err := json.Unmarshal(data, &cfg); err != nil {
+		return nil, err
+	}
+
+	for _, m := range cfg.Models {
+		c.ds.GetOrCreateModel(m.Name)
```
Mohammad-nassar10, line 194: A config entry `{"name":""}` will call `GetOrCreateModel("")` and register an empty-string key in the datastore silently. Suggest skipping empty names.

pkg/framework/plugins/datalayer/modelconfigcollector/plugin.go
```diff
@@ -0,0 +1,101 @@
...
+	data, err := os.ReadFile(c.modelsPath)
+	if err != nil {
+		return nil, err
+	}
+
+	var cfg ModelsConfig
+	if err := json.Unmarshal(data, &cfg); err != nil {
+		return nil, err
+	}
+
```
Mohammad-nassar10, line 186: Poll is additive-only, if a model is removed from the config file it stays in the datastore. Is this intentional or should we handle this?

pkg/framework/plugins/datalayer/modelconfigcollector/plugin_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/datastore"
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/datalayer"
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/plugin"
+	ctrlbuilder "sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// fakeHandle implements plugin.Handle for unit tests.
+type fakeHandle struct{ ds datalayer.Datastore }
```
Mohammad-nassar10, line 35: Another `fakeHandle` exists in the code (pkg/framework/plugins/datalayer/requestmetadata/plugin_test.go). Should we add this to a shared test utility?

ronenkat: Each test have different requirements from the fakehandle.
And the fake handle would end up the same as the actual Handle.

pkg/framework/plugins/datalayer/modelconfigcollector/plugin.go
ronenkat: Added removal code as well.

pkg/framework/plugins/datalayer/modelconfigcollector/plugin.go
ronenkat: added additional defensive programming code.

Mohammad-nassar10 (approved):

ronenkat:
Should be merged after #154

pkg/framework/plugins/datalayer/modelconfigcollector/plugin.go
```diff
@@ -0,0 +1,205 @@
...
+				absEvent, err := filepath.Abs(event.Name)
+				if err != nil {
+					logger.Error(err, "failed to resolve event path", "path", event.Name)
+					continue
+				}
+				absTarget, err := filepath.Abs(c.modelsPath)
+				if err != nil {
+					logger.Error(err, "failed to resolve target path", "path", c.modelsPath)
+					continue
+				}
```
shmuelk: Do this outside the goroutine and pass it in as a parameter.

ronenkat: done.

Mohammad-nassar10 (approved):

PR #131, opened 2026-05-26
Config part 3

shmuelk:
**What type of PR is this?**
/kind feature

This PR parses more of the text based configuration. In particular it loads the profiles from the configuration file.

In addition the old --plugin command line argument has been removed and instead the RequestProcessors and ReponseProcessors are taken from the profile in the configuration.

**NOTE:** At this time only one profile in the configuration is supported by the runner at this time.

Refs: #77 

```release-note
The --plugin command line argument has been removed and instead one should use the new text based configuration
```


ronenkat (approved):
Nice.

apix/config/v1alpha1/payloadprocessorconfig_types.go
```diff
@@ -34,16 +34,47 @@ type PayloadProcessorConfig struct {
 	// +kubebuilder:validation:Required
 	// Plugins is the list of plugins that will be instantiated.
 	Plugins []PluginSpec `json:"plugins"`
+
+	// +optional
+	// Preprocessing is an optional ordered list of references to plugins
+	// that will pre-process incoming requests
+	PreProcessing *PluginRefList `json:"preProcessing"`
+
+	// *optional
```
nirrozenbaum:
```suggestion
	// +optional
```

apix/config/v1alpha1/payloadprocessorconfig_types.go
```diff
@@ -34,16 +34,47 @@ type PayloadProcessorConfig struct {
...
+	// ProfilePicker is the plugin that chooses the profile to be used
+	ProfilePicker *PluginRef `json:"profilePicker"`
+
+	// +required
+	// +kubebuilder:validation:Required
+	// Profiles is the list of named Profiles
+	// that will be created.
+	Profiles []Profile `json:"profiles"`
```
nirrozenbaum, line 59: we can use kubebuilder validation to make sure at least one profile is defined.
```suggestion
	// +required
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MinItems=1
	// Profiles is the list of named Profiles
	// that will be created.
	Profiles []Profile `json:"profiles"`
```

apix/config/v1alpha1/payloadprocessorconfig_types.go
```diff
@@ -77,3 +108,81 @@ func (ps PluginSpec) String() string {
...
 	return "{" + strings.Join(parts, ", ") + "}"
 }
+
+// PluginRefList is a list of references to plugins
+type PluginRefList struct {
+	// +required
+	// +kubebuilder:validation:Required
+	// Plugins is the list of plugins being referenced.
+	Plugins []PluginRef `json:"plugins"`
+}
```
nirrozenbaum, line 126: ditto, make sure at least one otherwise there's no meaning for the `required` option.
```suggestion
	// +required
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MinItems=1
	// Plugins is the list of plugins being referenced.
	Plugins []PluginRef `json:"plugins"`
}
```

apix/config/v1alpha1/payloadprocessorconfig_types.go
```diff
@@ -77,3 +108,81 @@ func (ps PluginSpec) String() string {
...
+
+func (prl PluginRefList) String() string {
+	contents := ""
+	if len(prl.Plugins) > 0 {
```
nirrozenbaum, line 130: and if we add `minItems=1`, this conditional is not needed.

apix/config/v1alpha1/payloadprocessorconfig_types.go
```diff
@@ -77,3 +108,81 @@ func (ps PluginSpec) String() string {
...
+}
+
+// PluginRef is a reference to an instantiated plugin by name.
+type PluginRef struct {
+	// +required
+	// +kubebuilder:validation:Required
+	// PluginRef specifies a particular Plugin instance. The reference
+	// is to the name of an entry of the Plugins defined in the
+	// configuration's Plugins section.
+	PluginRef string `json:"pluginRef"`
```
nirrozenbaum, line 192: avoid empty string in required field:

```suggestion
type PluginRef struct {
	// +required
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MinLength=1
	// PluginRef specifies a particular Plugin instance. The reference
	// is to the name of an entry of the Plugins defined in the
	// configuration's Plugins section.
	PluginRef string `json:"pluginRef"`
```

apix/config/v1alpha1/payloadprocessorconfig_types.go
```diff
@@ -77,3 +108,81 @@ func (ps PluginSpec) String() string {
...
+	return fmt.Sprintf("{%s}", contents)
+}
+
+// Profile contains the information to create a Profile
+// entry to be used by the PayloadProcessor.
+type Profile struct {
+	// +required
+	// +kubebuilder:validation:Required
+	// Name specifies the name of this PayloadProcessorProfile
+	Name string `json:"name"`
```
nirrozenbaum, line 143: avoid empty name in required field:

```suggestion
	// +required
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MinLength=1
	// Name specifies the name of this PayloadProcessorProfile
	Name string `json:"name"`
```

apix/config/v1alpha1/payloadprocessorconfig_types.go
```diff
@@ -34,16 +34,47 @@ type PayloadProcessorConfig struct {
 	// +kubebuilder:validation:Required
```
nirrozenbaum, line 40:
```suggestion
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MinItems=1
```

apix/config/v1alpha1/payloadprocessorconfig_types.go
```diff
@@ -34,16 +34,47 @@ type PayloadProcessorConfig struct {
...
-		parts = append(parts, fmt.Sprintf("Plugins: %v", cfg.Plugins))
+	fmt.Fprintf(contents, "Plugins: %v", cfg.Plugins)
+
+	if cfg.PreProcessing != nil {
+		fmt.Fprintf(contents, ", PreProcessing: %v", cfg.PreProcessing)
+	}
+	if cfg.ProfilePicker != nil {
+		fmt.Fprintf(contents, ", ProfilePicker: %v", cfg.ProfilePicker)
+	}
+	if len(cfg.Profiles) > 0 {
```
nirrozenbaum: use kubebuilder validation. len(profiles) must be larger than 0 (so this if is not needed)

apix/config/v1alpha1/payloadprocessorconfig_types.go
```diff
@@ -77,3 +108,81 @@ func (ps PluginSpec) String() string {
...
+type Profile struct {
+	// +required
+	// +kubebuilder:validation:Required
+	// Name specifies the name of this PayloadProcessorProfile
+	Name string `json:"name"`
+
+	// +required
+	// +kubebuilder:validation:Required
+	// Plugins is the list of plugins for this Profile.
+	Plugins ProfilePlugins `json:"plugins"`
```
nirrozenbaum: do we check somewhere if plugins is empty?

pkg/config/loader/configloader.go
```diff
@@ -97,3 +109,46 @@ func instantiatePlugins(configuredPlugins []configapi.PluginSpec, handle plugin.
...
 	return nil
 }
+
+func buildProfiles(rawProfiles []configapi.Profile, handle plugin.Handle) (map[string]requesthandling.Profile, error) {
+	profiles := map[string]requesthandling.Profile{}
+
+	for _, rawProfile := range rawProfiles {
+		if len(rawProfile.Name) == 0 {
+			return nil, errors.New("a profile was specified without a name")
+		}
```
nirrozenbaum, line 127: if we do this validation using kubebuilder, this can't happen

pkg/config/loader/configloader.go
```diff
@@ -97,3 +109,46 @@ func instantiatePlugins(configuredPlugins []configapi.PluginSpec, handle plugin.
...
+				return nil, fmt.Errorf("there is no plugin named %s", pluginRef.PluginRef)
+			}
+			thePlugin, ok := rawPlugin.(requesthandling.ResponseProcessor)
+			if !ok {
+				return nil, fmt.Errorf("the plugin named %s is not a ResponseProcessor", pluginRef.PluginRef)
+			}
+			theProfile.ResponsePlugins[idx] = thePlugin
+		}
+
+		profiles[rawProfile.Name] = theProfile
```
nirrozenbaum, line 164: since both request/response plugins are optional, would here be the best place to check the profile has any plugins? 
or alternatively, should it be possible to define a profile without any plugin?

pkg/framework/interface/requesthandling/plugins.go
```diff
@@ -22,13 +22,34 @@ import (
 	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/plugin"
 )
```
nirrozenbaum, line 24: nice

pkg/server/options.go
```diff
@@ -93,7 +91,6 @@ func (opts *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&opts.EnablePprof, "enable-pprof", opts.EnablePprof,
 		"Enables pprof handlers. Defaults to true. Set to false to disable pprof handlers.")
 
-	fs.Var(&opts.PluginSpecs, "plugin", `Repeatable. --plugin <type>:<name>[:<json>]`)
```
nirrozenbaum, line 96: can we remove the option also from the helm chart and add a way to configure plugins through helm?
after this change, there's no way to configure plugins through helm deployment.

apix/config/v1alpha1/payloadprocessorconfig_types.go
shmuelk: An emperical test, shows that this does not work the way you wanted. Here is a log message from the IPP:

{"level":"info","ts":1779800563.451228,"logger":"setup","caller":"loader/configloader.go:74","msg":"Loaded raw configuration","config":"{Plugins: [{Type: body-field-to-header, Parameters: {\"fieldName\":\"model\",\"headerName\":\"X-Gateway-Model-Name\"}}], Profiles: [{Name: , Plugins: {Request: [{PluginRef: body-field-to-header}]}}]}"}

Note the `Name:,` in the text.

What adding `+kubebuilder:validation:MinLength=1` does is add JSON Schema to the generated CRD. We don't have CRDs here....

shmuelk:
To followup none of the Kubernetes validation tags actually work in the case of the pseudo CR configuration file as there is no CRD with a JSON Schema generated.

I'm not sure it's worth trying to generate and load the CRD and somehow extract the JSON Schema for use in validation.

apix/config/v1alpha1/payloadprocessorconfig_types.go
```diff
@@ -32,18 +38,51 @@ type PayloadProcessorConfig struct {
...
 
-	if len(cfg.Plugins) > 0 {
-		parts = append(parts, fmt.Sprintf("Plugins: %v", cfg.Plugins))
+	if cfg.PreProcessing != nil {
+		fmt.Fprintf(contents, ", PreProcessing: %v", cfg.PreProcessing)
+	}
+	if cfg.ProfilePicker != nil {
+		fmt.Fprintf(contents, ", ProfilePicker: %v", cfg.ProfilePicker)
+	}
+	if len(cfg.Profiles) > 0 {
```
nirrozenbaum: nit: might be good to remove the "if" as we want to see empty profiles in the log (will help debug in error cases)

config/charts/payload-processor/templates/config.yaml
```diff
@@ -0,0 +1,27 @@
...
+        headerName: X-Gateway-Model-Name
+    - type: base-model-to-header
+    profiles:
+    - name: default
+      plugins:
+        request:
+        - pluginRef: body-field-to-header
+        - pluginRef: base-model-to-header
+    {{- end }}
+---
```
nirrozenbaum: pls add extra line

nirrozenbaum: in this setup we will have either the default plugins OR custom plugins. 
what are your thoughts about having both options baked into the config map allowing to easily switch between configurations and compare default vs custom?
example here (from previous releases of epp):
https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/release-1.2/config/charts/inferencepool/templates/epp-config.yaml#L1

nirrozenbaum (commented):
@shmuelk thanks. the PR looks good. 
I left one question about the helm chart config (if we want to bake a single config or allowing default + custom config). 
if we allow two configs, we should also add to the template the selected config (and not use hardcoded ipp-config.yaml in the deplyment template). 

also, would it be possible to update the sample (under deploy) to include all possible fields? current sample shows request plugins but not response, and also not showing multiple profiles and profile picker.

shmuelk:
@nirrozenbaum wrote:

>also, would it be possible to update the sample (under deploy) to include all possible fields? current sample shows request plugins but not response, and also not showing multiple profiles and profile picker.

WRT to adding a response entry, there are no sample ResponseProcessing plugins in the code base. If someone has an idea for one, I can happily change things here.

The current code as stated in the PR description, does not yet support the profile picker and such supports only one profile at this time. The rest of the code base does not support a ProfilePicker yet. I added the definitions, but that's it.

nirrozenbaum:
/lgtm
/approve

Thanks!

PR #124, opened 2026-05-20
Adds tracking of actual token usage per model.

davidbreitgand:
## What does this PR do?

1. Adds tracking of actual token usage per model
2. Hardens implementation: protects against an accidental writing to the `nil` datastore 
3. Current implementation is limited to OpenAI as Phase 1, other providers will be added in subsequent PRs 

## Why is this change needed?

requestmetadata/plugin.go extracts max_tokens on the inbound request event. max_tokens is optional and is only a proxy for the actual token usage. The actual token usage will be instrumental for advanced cost-aware scoring, accurate cost tracking to capture price reversals, cache management, and visualization.

## How was this tested?

- [x] Unit tests added/updated
- [ ] Integration/e2e tests added/updated
- [ ] Manual testing performed

## Checklist

- [x] Commits are signed off (`git commit -s`) per [DCO](PR_SIGNOFF.md)
- [x] Code follows project [contributing guidelines](CONTRIBUTING.md)
- [x] Tests pass locally (`make test`)
- [x] Linters pass (`make lint`)
- [ ] Documentation updated (if applicable)

## Related Issues

Fixes #115

PR #132, opened 2026-05-26
Expected wait scorer proposal.

Mohammad-nassar10:
**What type of PR is this?**

kind feature

Adding a scorer that takes into account the expected waiting time to process the request in each model.

*Note*: Depends on #124, which populates `TokensPerSec` in `RequestMetadataExtractor`.

Fixes #

```release-note
NONE
```

### DIFF
--- a/docs/proposals/expected-wait-scorer/README.md
+++ b/docs/proposals/expected-wait-scorer/README.md
@@ -1,0 +1,130 @@
+# Expected-Wait Model Scorer
+
+Author(s): @Mohammad-nassar10
+
+
+## Summary
+
+The `InflightRequestsScorer` routes to the least-loaded model by request count alone. This ignores model speed and request weight. A model with 100 queued requests at 1000 tokens/sec is a better choice than one with 10 queued requests at 10 tokens/sec — yet the current scorer picks the wrong one.
+
+This proposal introduces an **`ExpectedWaitScorer`** that estimates the total response time for the current request on each candidate model and picks the shortest. It also covers how per-model output throughput is tracked.
+
+
+---
+
+## Options Considered
+
+### Option A — Request-count × average latency (baseline)
+
+**Formula:** `expected_wait = inflight_requests × avg_latency_ms`
+
+**Data needed:**
+- `RequestMetadataCount.Requests` (existing extractor)
+- `avg_latency_ms` — running mean of `p.Duration`, tracked per model
+
+**Pros:** Simple. No changes to existing extractors.
+
+**Cons:** Treats a 10-token request and a 10,000-token request as equal queue weight.
+
+---
+
+### Option B — Inflight-tokens ÷ tokens-per-second (token-aware)
+
+**Formula:** `total_time = (inflight_tokens + max_tokens) / tokens_per_sec`
+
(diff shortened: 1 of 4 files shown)


docs/proposals/expected-wait-scorer/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Options Considered
+
+### Option A — Request-count × average latency (baseline)
+
+**Formula:** `expected_wait = inflight_requests × avg_latency_ms`
+
+**Data needed:**
+- `RequestMetadataCount.Requests` (existing extractor)
+- `avg_latency_ms` — running mean of `p.Duration`, tracked per model
```
aradhalevy, line 23: Is the total Duration calculated from the moment the requests arrives to the system? from the moment is arrives to the model? or from the moment the vLLM pod starts to process it?

docs/proposals/expected-wait-scorer/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+**Pros:** Simple. No changes to existing extractors.
+
+**Cons:** Treats a 10-token request and a 10,000-token request as equal queue weight.
+
+---
+
+### Option B — Inflight-tokens ÷ tokens-per-second (token-aware)
+
+**Formula:** `total_time = (inflight_tokens + max_tokens) / tokens_per_sec`
```
aradhalevy, line 33: I think the number of Input tokens could also be an indicator (though less meaningful than avg output tokens in next steps). WDYT?

docs/proposals/expected-wait-scorer/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+### New Types
+
+**`pkg/framework/plugins/modelselector/scorer/expectedwait`**
+
+```go
+const PluginType = "expected-wait-scorer"
+```
+
+`ExpectedWaitScorer.Score` reads the `"request-metadata"` attribute via `datalayer.ReadAttributeKey[T]` and min-max normalises. Models with no throughput data (`tokensPerSec == 0`) receive `expectedWait = 0` — optimistic, so untried models are preferred.
```
aradhalevy, line 88: As we saw in our benchmark, some data about the actual system load could take time to arrive to the scorers. In an edge case where there are a lot of requests at once and you spin up a new InferencePool, it might be overloaded with many requests as it gets the "best" score. It is only an edge case though

docs/proposals/expected-wait-scorer/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+The correct proxy is the model's observed **average `completion_tokens`** per response, which `RequestMetadataExtractor` can track alongside throughput:
+
+```go
+type RequestMetadataCount struct {
+    Requests            int64
+    Tokens              int64
+    OutputTokens        int64
+    TokensPerSec        float64
+    AvgCompletionTokens float64  // rolling average of usage.completion_tokens
```
ronenkat, line 115: Let average completion tokens by request size bin, e.g., for a given request size bin the average completion tokens for the bin.

ronenkat (approved):

PR #136, opened 2026-05-27
[Feature] wire the datalayer plugins into the configuration 

### LINKED ISSUE #135: [Feature] Wire the data layer plugins into the configuration and runner
Add configuration support for `NotificationSource` data layer plugins. Currently a
`NotificationSource` is hardcoded in `cmd/runner/runner.go` with a fixed extractor
(`RequestMetadataExtractor`), and there is no way to declare or customize notification
sources through the configuration file.

Users should be able to declare `NotificationSource` plugins in the `plugins` section
and reference them in a new top-level `notificationSources` field, the same way profiles
reference plugins today. The runner should start all configured sources as part of the
manager lifecycle (ensuring `Stop()` is called on shutdown).

**Why is this needed**:
The hardcoded path is inflexible and prevents users from configuring the data pipeline
through the standard YAML config file. It also has a bug: `Stop()` is never called on
shutdown, so the notification source goroutine is not cleanly drained. Making this
config-driven aligns with the existing plugin/profile pattern and fixes the lifecycle gap.

ronenkat:
**What type of PR is this?**
/kind feature

Replaces the hardcoded `NotificationSource` in `cmd/runner/runner.go` with a
config-driven approach. Users now declare notification-source plugins in the `plugins`
section and list them under a new top-level `notificationSources` field in the
`PayloadProcessorConfig`. The runner starts each configured source via the
controller-runtime manager, which guarantees `Stop()` is called on graceful shutdown.

Changes:
- `apix/config/v1alpha1`: added `NotificationSources []PluginRef` field to
  `PayloadProcessorConfig` (and updated generated DeepCopy)
- `pkg/config`: added `NotificationSources []datasource.NotificationSource` to `Config`
- `pkg/config/loader`: added `buildDatalayer` called after `buildProfiles`; resolves
  plugin refs and type-asserts each to `NotificationSource`
- `internal/runnable`: new `DataSourceRunnable` adapter that starts a `DataSource`,
  blocks until context cancellation, then calls `Stop()`
- `cmd/runner`: `loadConfiguration` now returns `*config.Config`; replaces hardcoded
  block with a loop that registers each source with the manager
- `deploy/config/ipp-datalayer-config.yaml`: example config wiring a notification-source
  with a request-metadata extractor

Fixes #135 

```release-note
NotificationSource plugins can now be declared in the IPP configuration file and are
started/stopped automatically as part of the manager lifecycle.

### CODE REVIEW

File: internal/runnable/datasource.go [hunk 1]
```diff
@@ -0,0 +1,38 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/datalayer/datasource"
+)
+
+// DataSourceRunnable wraps a DataSource as a manager.Runnable.
+// It starts the data source, waits for context cancellation, then stops it gracefully.
+func DataSourceRunnable(ds datasource.DataSource) manager.Runnable {
+	return manager.RunnableFunc(func(ctx context.Context) error {
```
shmuelk: This code is alreday running as a goroutine. Why does the called Start() function launch another one?

ronenkat: Moving to use start() and stop() directly, no need for fancy management.

apix/config/v1alpha1/payloadprocessorconfig_types.go
```diff
@@ -75,10 +79,17 @@ func (cfg PayloadProcessorConfig) String() string {
 	if cfg.ProfilePicker != nil {
 		fmt.Fprintf(contents, ", ProfilePicker: %v", cfg.ProfilePicker)
 	}
+<<<<<<< HEAD
+
+=======
+>>>>>>> add-notification-source-config
```
shmuelk: This couldn't have compiled....

ronenkat: Done.

PR #138, opened 2026-05-28
feat: parse SSE streaming responses for response plugins

noyitz:
## Summary

- Parse SSE (Server-Sent Events) streaming responses for response plugins when JSON parsing fails
- Fix streaming response header handling: always respond so Envoy proceeds with body chunks
- Fix streaming response body chunk handling: send immediate ack for non-EoS chunks

## Problem

When providers like Anthropic return streaming responses (SSE format with `text/event-stream`), the response body starts with `event:` not `{`. The current code fails JSON parsing and skips all response plugins:

```
Failed to parse response body as JSON, skipping response plugins
```

Usage-tracking, metering, and any other response plugins never execute for streaming requests. Additionally, `HandleResponseHeaders` returns `nil` for streaming, causing Envoy `per-message_timeout_exceeded`.

## Solution

**`pkg/handlers/response.go`:**
1. When JSON parsing fails, try SSE parsing — scan `data:` lines for JSON objects, extract `usage` and `model` fields
2. Always respond to response headers so Envoy sends body chunks

**`pkg/handlers/server.go`:**
3. Send immediate `BodyResponse{}` ack for non-EoS response body chunks so Envoy continues streaming

Client sees streamed output in real-time. Chunks accumulated in-memory (bounded by max_tokens) and parsed at EoS.

## Test plan
- [x] Anthropic streaming API (claude-opus-4-6, stream:true) — SSE usage extracted
- [x] Response plugins execute for streaming responses
- [x] Non-streaming responses unchanged
- [x] Usage data recorded in PostgreSQL via metering service

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/pkg/handlers/response.go
+++ b/pkg/handlers/response.go
@@ -17,6 +17,7 @@ limitations under the License.
 package handlers
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -44,9 +45,9 @@ func (s *Server) HandleResponseHeaders(ctx context.Context, reqCtx *RequestConte
 
 	if !headers.GetEndOfStream() {
 		log.FromContext(ctx).V(logutil.VERBOSE).Info("captured response headers, deferring response until body arrives...")
-		return nil
 	}
-	// EndOfStream means no body is expected, return HeadersResponse immediately
+	// Always respond to response headers so Envoy proceeds with body chunks.
+	// In STREAMED/FULL_DUPLEX_STREAMED mode, Envoy blocks until we respond.
 	return []*eppb.ProcessingResponse{
 		{
 			Response: &eppb.ProcessingResponse_ResponseHeaders{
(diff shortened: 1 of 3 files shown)


pkg/handlers/response.go
```diff
@@ -64,8 +65,15 @@ func (s *Server) HandleResponseBody(ctx context.Context, reqCtx *RequestContext,
...
+		// Try parsing as SSE (Server-Sent Events) — streaming responses from providers
+		// like Anthropic use SSE format which isn't valid JSON.
+		if sseBody, sseErr := parseSSEResponseBody(responseBodyBytes); sseErr == nil && sseBody != nil {
+			reqCtx.Response.Body = sseBody
+			logger.V(logutil.VERBOSE).Info("parsed SSE response body for response plugins")
+		} else {
+			logger.Error(err, "Failed to parse response body as JSON or SSE, skipping response plugins")
+			return s.generateEmptyResponseBodyResponse(responseBodyBytes), nil
+		}
 	}
```
shmuelk: Why try to parse a JSON and fail? Why not look at the content-type header and parse accordingly?

noyitz:
Closing — with the body mode framework (PR #169) and ChunkProcessor (PR #170), we always use FULL_DUPLEX_STREAMED mode. SSE responses are handled by chunk processors in the plugin layer, not by the framework's HandleResponseBody. The buffered SSE parsing path is no longer needed.

PR #142, opened 2026-05-31
[Cleanup] Model select profile to pipeline rename

ronenkat:
**What type of PR is this?**
/kind cleanup

Diffrenciate between IPP profiles and model-selector pipelines.

Follow-up on #127 

```release-note
NONE
```


nirrozenbaum:
should be reviewed after #127 merges

PR #143, opened 2026-05-31
Add event notification hooks for request and response.

Mohammad-nassar10:
**What type of PR is this?**
/kind feature

Integrate notification source with the request and response flows.

Fixes #


pkg/handlers/request.go
```diff
@@ -71,6 +72,14 @@ func (s *Server) HandleRequestBody(ctx context.Context, reqCtx *RequestContext,
 		return nil, err
 	}
 
+	// Notify the data layer of the incoming request.
```
ronenkat: lets move after line 95 as the request headers are updated below.

pkg/handlers/response.go
```diff
@@ -58,6 +59,20 @@ func (s *Server) HandleResponseHeaders(ctx context.Context, reqCtx *RequestConte
 
 // HandleResponseBody handles response bodies by executing response plugins in order.
 func (s *Server) HandleResponseBody(ctx context.Context, reqCtx *RequestContext, responseBodyBytes []byte) ([]*eppb.ProcessingResponse, error) {
+	reqCtx.ResponseCompleteTimestamp = time.Now()
```
ronenkat: We should capture the timestamps in `server.go`, in `case *extProcPb.ProcessingRequest_RequestBody:`.
We can capture the first response body message as TTFT proxy timestamp and after `if !v.RequestBody.EndOfStream` we can capture the full end-to-end time.
Suggest you extract to a separate pull-request

cmd/runner/runner.go
```diff
@@ -215,6 +215,9 @@ func (r *Runner) Run(ctx context.Context) error {
 		ProfilePicker: r.profilePicker,
 		Profiles:      r.profiles,
 	}
+	if len(r.notificationSources) > 0 {
+		serverRunner.EventNotifier = r.notificationSources[0]
+	}
```
shmuelk: PR #154 removed notificationSources from the Runner struct

pkg/handlers/server.go
```diff
@@ -54,16 +55,24 @@ func NewServer(profilePicker requesthandling.ProfilePicker, profiles map[string]
...
+	s.eventNotifier = n
+	return s
+}
+
 // Server implements the Envoy external processing server.
 // https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/ext_proc/v3/external_processor.proto
 type Server struct {
 	profilePicker requesthandling.ProfilePicker
 	profiles      map[string]*requesthandling.Profile
+  eventNotifier   datasource.EventNotifier
```
shmuelk: Format please

pkg/server/runserver.go
```diff
@@ -39,6 +40,7 @@ type ExtProcServerRunner struct {
 	SecureServing bool
 	ProfilePicker requesthandling.ProfilePicker
 	Profiles      map[string]*requesthandling.Profile
+  EventNotifier   datasource.EventNotifier
```
shmuelk: Format please

pkg/server/runserver.go
```diff
@@ -67,7 +69,11 @@ func (r *ExtProcServerRunner) AsRunnable(logger logr.Logger) manager.Runnable {
 			srv = grpc.NewServer()
 		}
 
-		extProcPb.RegisterExternalProcessorServer(srv, handlers.NewServer(r.ProfilePicker, r.Profiles))
+		server := handlers.NewServer(r.ProfilePicker, r.Profiles)
+		if r.EventNotifier != nil {
```
shmuelk: I see that you have added to the ExtProcServerRunner an EventNotifier field. 

What piece of code sets it?

pkg/handlers/server.go
```diff
@@ -54,16 +55,24 @@ func NewServer(profilePicker requesthandling.ProfilePicker, profiles map[string]
 	}
 }
 
+// WithEventNotifier sets the event notifier used to feed the data layer.
+func (s *Server) WithEventNotifier(n datasource.EventNotifier) *Server {
+	s.eventNotifier = n
+	return s
+}
+
```
shmuelk, line 63: When is this called?

pkg/handlers/request.go
```diff
@@ -90,7 +91,16 @@ func (s *Server) HandleRequestBody(ctx context.Context, reqCtx *RequestContext,
 		reqCtx.Request.SetHeader(contentLengthHeader, strconv.Itoa(len(requestBodyBytes)))
 	}
 
+	// Notify the data layer of the incoming request after headers are fully formed.
+	if s.eventNotifier != nil {
```
shmuelk: I understand that in the current code in this PR s.eventNotifier will always be nil, but in a more complete system shouldn't it always be not nil? These are built-in events that are being fired, shouldn't there always be. notifier?

This is of course unless the notifier comes from a configured plugin, which might be missing in some configurations.

Mohammad-nassar10:
Replaced by #158.

pkg/server/runserver.go
Mohammad-nassar10: It was missing, added it to the runner (`cmd/runner/runner.go`) to pass `r.processor`, which already implements `datasource.EventNotifier`. The fix on #158.

pkg/handlers/server.go
Mohammad-nassar10: From `ExtProcServerRunner.AsRunnable` on every server start.

pkg/handlers/request.go
Mohammad-nassar10: Agreed, the runner unconditionally wires `r.processor` now, so the nil guards were dead code. Dropped them in both `request.go` and `response.go`. Tests use a tiny no-op notifier.
Fixed in #158.

PR #144, opened 2026-05-31
added single profile picker

nirrozenbaum:
**What type of PR is this?**
/kind feature

add `single-profile-picker`. this is the first profile picker that's being added and is still not used. 
as a follow up, config loader should use it as default value when no profile picker is explicitly defined and a single profile is configured.

Fixes #

```release-note
NONE
```


nirrozenbaum:
@yehuditkerido from some reason the workflow to tag the PR kind is adding all kinds (this is something new). 
can you look into it?
removing now the kind labels manually, but if you scroll in the PR history you can easily see it.

PR #145, opened 2026-06-01
fix: skip HTML comments when parsing /kind labels

yehuditkerido:
**What type of PR is this?**
/kind bug

Fixes the `pr-kind-labeler` workflow that was adding all `kind/*` labels to every PR.

**Root cause:** `grep '^/kind '` matched `/kind` lines inside HTML comments in the PR template.

**Fix:** Use `awk` with state tracking to skip HTML comments.

Fixes issue reported in https://github.com/llm-d/llm-d-inference-payload-processor/pull/144#issuecomment-4587812938

```release-note
NONE

### DIFF
--- a/.github/workflows/pr-kind-labeler.yaml
+++ b/.github/workflows/pr-kind-labeler.yaml
@@ -27,8 +27,12 @@ jobs:
             done
 
           # Extract /kind values from PR body and apply labels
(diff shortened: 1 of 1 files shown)


nirrozenbaum:
/lgtm
/approve

Thanks!

PR #146, opened 2026-06-01
Capture TTFT and E2E timestamps.

Mohammad-nassar10:
**What type of PR is this?**
/kind feature

Capture TTFT and E2E timestamps in response body processing loop. This will be used in the `requestmetadata` extractor.

Fixes #

### DIFF
--- a/pkg/handlers/server.go
+++ b/pkg/handlers/server.go
@@ -63,6 +63,7 @@ type Server struct {
 // RequestContext stores context information during the lifetime of an HTTP request.
 type RequestContext struct {
 	RequestReceivedTimestamp  time.Time
+	FirstChunkTimestamp       time.Time
 	ResponseCompleteTimestamp time.Time
 	CycleState                *plugin.CycleState
 	Request                   *requesthandling.InferenceRequest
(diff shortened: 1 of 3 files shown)


pkg/handlers/server.go
```diff
@@ -63,6 +63,7 @@ type Server struct {
 // RequestContext stores context information during the lifetime of an HTTP request.
 type RequestContext struct {
 	RequestReceivedTimestamp  time.Time
+	FirstChunkTimestamp       time.Time
```
nirrozenbaum: nit: request is also sent in chunks, so just to be super clear I would prefix it with Response.

```suggestion
	ResponseFirstChunkTimestamp       time.Time
```

nirrozenbaum (commented):
@Mohammad-nassar10 nice, looks good. 
would it be possible to add a metric exposing TTFT with label `model` (and if additional labels are needed we can add as well)?

Mohammad-nassar10:
@nirrozenbaum yes sure, added a metric for TTFT, thanks.

nirrozenbaum:
/lgtm
/approve

nice!

PR #148, opened 2026-06-01
test(modelselector): add unit tests for random-picker plugin.

### LINKED ISSUE #147: Add unit tests for random-picker plugin
The `random-picker` plugin (`pkg/framework/plugins/modelselector/picker/random`) is currently the only Picker plugin without unit tests. Both `maxscore` and `weightedrandom` pickers have test coverage.

### Acceptance criteria
- Add `picker_test.go` with tests for all public functions
- Use table-driven test format where appropriate
- Verify the plugin's unique behavior: uniform random selection that ignores scores

### References
- `pkg/framework/plugins/modelselector/picker/maxscore/picker_test.go`
- `pkg/framework/plugins/modelselector/picker/weightedrandom/picker_test.go`

yehuditkerido:
**What type of PR is this?**
/kind test

Add unit tests for the `random-picker` plugin, which was the only Picker plugin without test coverage. Tests use table-driven format and verify all public functions including a statistical test that confirms scores are ignored and selection is uniformly distributed.

Fixes #147 

```release-note
NONE

### DIFF
--- a/pkg/framework/plugins/modelselector/picker/random/picker_test.go
+++ b/pkg/framework/plugins/modelselector/picker/random/picker_test.go
@@ -1,0 +1,192 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package random
+
+import (
+	"context"
+	"testing"
+
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/datalayer"
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/modelselector"
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/plugin"
+)
+
+func TestRandomPicker_Pick(t *testing.T) {
+	modelA := datalayer.NewModel("model-a")
+	modelB := datalayer.NewModel("model-b")
+	modelC := datalayer.NewModel("model-c")
+
+	tests := []struct {
+		name  string
+		input []*modelselector.ScoredModel
+	}{
+		{
+			name: "single model returns that model",
+			input: []*modelselector.ScoredModel{
+				{Model: modelA, Score: 1.0},
+			},
+		},
+		{
+			name: "returns a model from multiple candidates",
+			input: []*modelselector.ScoredModel{
+				{Model: modelA, Score: 0.1},
+				{Model: modelB, Score: 0.2},
+				{Model: modelC, Score: 0.7},
+			},
+		},
+		{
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: pkg/framework/plugins/modelselector/picker/random/picker_test.go [hunk 1]
```diff
@@ -0,0 +1,285 @@
...
+
+func TestRandomPicker_Pick(t *testing.T) {
+	modelA := datalayer.NewModel("model-a")
+	modelB := datalayer.NewModel("model-b")
+	modelC := datalayer.NewModel("model-c")
+
+	tests := []struct {
+		name      string
+		input     []*modelselector.ScoredModel
+		wantModel string // empty means any model from input is valid
```
nirrozenbaum: you have in the test itself a section saying:
```
// Verify returned model is one of the inputs
```

and `wantModel` is used here only when there's a single model.. looks like this field can be removed

pkg/framework/plugins/modelselector/picker/random/picker_test.go
```diff
@@ -0,0 +1,285 @@
...
+					t.Errorf("model %q selected %d times (%.1f%%), expected between %d%% and %d%%",
+						name, count, float64(count)/float64(tt.iterations)*100,
+						tt.minPercent, tt.maxPercent)
+				}
+			}
+		})
+	}
+}
+
+func TestRandomPicker_Pick_AllModelsReachable(t *testing.T) {
```
nirrozenbaum: isn't this tested already in previous test? (although it's only with 2 models, but yet)

pkg/framework/plugins/modelselector/picker/random/picker_test.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+}
+
+func TestRandomPicker_WithName(t *testing.T) {
+	tests := []struct {
+		name         string
+		newName      string
+		wantName     string
+		wantType     string
+		wantSameAddr bool
```
nirrozenbaum: this is always true.. so.. what is it testing?

nirrozenbaum (commented):
@yehuditkerido thanks, looks good. 
left few nits

pkg/framework/plugins/modelselector/picker/random/picker_test.go
yehuditkerido: Yes, I see what you're saying. I thought having `wantModel` would add clarity for the single-model case, but you're right — since the "verify returned model is one of the inputs" logic already covers it, the field is redundant. Will remove it.

pkg/framework/plugins/modelselector/picker/random/picker_test.go
yehuditkerido: You're right — if the uniform distribution test passes, it implicitly proves all models are reachable. I'll remove it to avoid redundancy.

pkg/framework/plugins/modelselector/picker/random/picker_test.go
```diff
@@ -0,0 +1,199 @@
...
+				{Model: modelA, Score: 1.0},
+			},
+		},
+		{
+			name: "returns a model from two candidates",
+			input: []*modelselector.ScoredModel{
+				{Model: modelA, Score: 0.5},
+				{Model: modelB, Score: 0.3},
+			},
+		},
```
nirrozenbaum: can we remove this test case?
covered by "three candidates"

```suggestion
```

nirrozenbaum:
/lgtm
/approve

Thanks!

PR #149, opened 2026-06-01
Configure the profile picker

shmuelk:
**What type of PR is this?**
/kind feature

This PR wires the profile picker specified in the configuration text to the IPP's configuration struct.

In addition, if the user did not specify a profile picker and only one profile has been specified a default profile picker (SingleProfilePicker) is added to the configuration.

!note If a profile picker plugin is referenced in the configuration's plugins section, it will be automatically used if one was not referenced in the configuration's profilePicker section.

?

```release-note
If a profile picker plugin is not mentioned in the configuration file and tonly one profile was specified, a default profile picker will be added to the configuration.
```

### DIFF
--- a/pkg/config/loader/defaults.go
+++ b/pkg/config/loader/defaults.go
@@ -73,9 +73,12 @@ func applyRawConfigDefaults(rawConfig *configapi.PayloadProcessorConfig) {
 func applyPluginDefaults(rawConfig *configapi.PayloadProcessorConfig, handle plugin.Handle) error {
 	if rawConfig.ProfilePicker == nil || rawConfig.ProfilePicker.PluginRef == "" {
 		var profilePicker requesthandling.ProfilePicker
-		var ok bool
 		for _, rawPlugin := range handle.GetAllPlugins() {
-			if profilePicker, ok = rawPlugin.(requesthandling.ProfilePicker); ok {
+			if aProfilePicker, ok := rawPlugin.(requesthandling.ProfilePicker); ok {
+				if profilePicker != nil {
+					return errors.New("multiple profile pickers have been defined in the configuration")
+				}
+				profilePicker = aProfilePicker
 				break
 			}
 		}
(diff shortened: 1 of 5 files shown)


pkg/config/loader/defaults.go
```diff
@@ -63,3 +68,60 @@ func applyRawConfigDefaults(rawConfig *configapi.PayloadProcessorConfig) {
...
+		}
+
+		if profilePicker == nil {
+			if len(rawConfig.Profiles) == 1 {
+				if err := registerDefaultPlugin(rawConfig, handle, single.SingleProfilePickerType); err != nil {
+					return err
+				}
+				profilePicker = handle.Plugin(single.SingleProfilePickerType).(requesthandling.ProfilePicker)
+			} else {
+				return errors.New("more than one profile in the configuration without a profile picker")
```
nirrozenbaum: a more user friendly err message?
```suggestion
				return errors.New("multiple profiles in the configuration require a profile picker")
```

pkg/config/loader/defaults.go
```diff
@@ -63,3 +68,60 @@ func applyRawConfigDefaults(rawConfig *configapi.PayloadProcessorConfig) {
...
+// applyPluginDefaults injects default plugins into the configuration
+func applyPluginDefaults(rawConfig *configapi.PayloadProcessorConfig, handle plugin.Handle) error {
+	if rawConfig.ProfilePicker == nil || rawConfig.ProfilePicker.PluginRef == "" {
+		var profilePicker requesthandling.ProfilePicker
+		var ok bool
+		for _, rawPlugin := range handle.GetAllPlugins() {
+			if profilePicker, ok = rawPlugin.(requesthandling.ProfilePicker); ok {
+				break
+			}
+		}
```
nirrozenbaum, line 83: do we check anywhere if there is more than one profile picker and return an error?
is here the right place to check?

shmuelk: Probably

shmuelk: Done

nirrozenbaum: I think current code still breaks when seeing the first profile picker, isn't it?

shmuelk: Fixed. The code now checks all of the defined plugins

PR #151, opened 2026-06-01
Add GetModels(predicate) to Datastore interface

### LINKED ISSUE #141: [Feature]: Optimize Datastore getModels interface for abstraction and performance

Refactor the `Datastore` interface to expose a `GetModels(predicate)` method that returns all models matching a given predicate in a single call, replacing the current two-step `Models()` + `GetOrCreateModel(name)` pattern.

Example signature:
```go
GetModels(predicate func(Model) bool) []Model
Callers that want all models pass a predicate that always returns true.

Why is this needed:

The current Datastore interface forces callers to loop over every model name and fetch each model individually:


// pkg/framework/plugins/requesthandling/modelselector/plugin.go
func (p *ModelSelectorPlugin) loadCandidateModels() []datalayer.Model {
    modelNames := p.datastore.Models()
    candidates := make([]datalayer.Model, len(modelNames))
    for i, name := range modelNames {
        candidates[i] = p.datastore.GetOrCreateModel(name)
    }
    return candidates
}
This pattern has two problems:

Wrong abstraction — the interface leaks its internal representation (a list of names) and forces the caller to re-assemble the full objects. The datastore already owns all the model data; it should be responsible for returning it.
Per-request overhead — loadCandidateModels is called on every inference request. The N individual GetOrCreateModel calls (and any locking/allocation they imply) happen on the hot path, where a single GetModels call could return a pre-built slice directly.
A GetModels(predicate) API fixes both: the interface is self-contained, filtering can happen inside the datastore (closer to the data), and callers get a clean single call on the hot path.

### PR DESCRIPTION
**Pull request by Jwrede**
## Summary

- Add `GetModels(predicate func(Model) bool) []Model` to the `Datastore` interface
- Implement in the in-memory store under a single read lock
- Update the test fake datastore to satisfy the new interface

Closes #141

## Details

The current `Datastore` interface forces callers to use a two-step pattern to retrieve models: call `Models()` to get a list of names, then loop over each name calling `GetOrCreateModel(name)` individually. This leaks the internal representation (a list of names) and imposes per-call locking overhead on the hot path.

`GetModels(predicate)` returns all matching models in a single call under one read lock. Callers that want all models pass a predicate that always returns true. The existing `Models()` and `GetOrCreateModel()` methods are unchanged for backward compatibility.

## Test plan

- [x] `TestGetModelsAll`: always-true predicate returns all models
- [x] `TestGetModelsWithPredicate`: attribute-based filtering returns correct subset
- [x] `TestGetModelsNoneMatch`: always-false predicate returns empty slice
- [x] `TestGetModelsEmpty`: empty store returns empty slice
- [x] All existing tests pass (12 tests in inmemory, full `./pkg/...` suite green)

### DIFF
--- a/pkg/datastore/inmemory/datastore_test.go
+++ b/pkg/datastore/inmemory/datastore_test.go
@@ -155,6 +155,74 @@ func TestAttributeNilValue(t *testing.T) {
 	}
 }
 
+// TestGetModelsAll tests that GetModels with an always-true predicate returns all models.
+func TestGetModelsAll(t *testing.T) {
+	s := NewDatastore()
+	s.GetOrCreateModel("gpt-4")
+	s.GetOrCreateModel("llama-3")
+	s.GetOrCreateModel("mistral")
+
+	models := s.GetModels(func(datalayer.Model) bool { return true })
+	if len(models) != 3 {
+		t.Errorf("expected 3 models, got %d", len(models))
+	}
+
+	names := make(map[string]bool)
+	for _, m := range models {
+		names[m.GetName()] = true
+	}
+	for _, expected := range []string{"gpt-4", "llama-3", "mistral"} {
+		if !names[expected] {
+			t.Errorf("missing expected model: %s", expected)
+		}
+	}
+}
+
+// TestGetModelsWithPredicate tests that GetModels filters correctly.
+func TestGetModelsWithPredicate(t *testing.T) {
+	s := NewDatastore()
+	s.GetOrCreateModel("gpt-4").GetAttributes().Put("vendor", testValue{Value: 1})
+	s.GetOrCreateModel("llama-3").GetAttributes().Put("vendor", testValue{Value: 2})
+	s.GetOrCreateModel("mistral").GetAttributes().Put("vendor", testValue{Value: 1})
+
+	models := s.GetModels(func(m datalayer.Model) bool {
+		v, ok := m.GetAttributes().Get("vendor")
+		return ok && v.(testValue).Value == 1
+	})
+	if len(models) != 2 {
+		t.Errorf("expected 2 models matching predicate, got %d", len(models))
+	}
+
+	names := make(map[string]bool)
+	for _, m := range models {
+		names[m.GetName()] = true
+	}
+	if !names["gpt-4"] || !names["mistral"] {
+		t.Errorf("expected gpt-4 and mistral, got %v", names)
+	}
+}
+
+// TestGetModelsNoneMatch tests that GetModels returns empty when no models match.
+func TestGetModelsNoneMatch(t *testing.T) {
+	s := NewDatastore()
+	s.GetOrCreateModel("gpt-4")
+	s.GetOrCreateModel("llama-3")
+
+	models := s.GetModels(func(datalayer.Model) bool { return false })
+	if len(models) != 0 {
+		t.Errorf("expected 0 models, got %d", len(models))
+	}
+}
+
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: pkg/datastore/inmemory/datastore_test.go [hunk 1]
```diff
@@ -155,6 +155,74 @@ func TestAttributeNilValue(t *testing.T) {
...
+		}
+	}
+}
+
+// TestGetModelsWithPredicate tests that GetModels filters correctly.
+func TestGetModelsWithPredicate(t *testing.T) {
+	s := NewDatastore()
+	s.GetOrCreateModel("gpt-4").GetAttributes().Put("vendor", testValue{Value: 1})
+	s.GetOrCreateModel("llama-3").GetAttributes().Put("vendor", testValue{Value: 2})
+	s.GetOrCreateModel("mistral").GetAttributes().Put("vendor", testValue{Value: 1})
```
nirrozenbaum, line 186: not related to this PR, but maybe it would be nicer to include PutAttribute/GetAttribute/DeleteAttribute functions directly on `Model` to avoid this weird line of `GetAttributes().Put`.

not a blocker of your PR, just an improvement suggestion as a follow up.

pkg/datastore/inmemory/datastore_test.go
```diff
@@ -155,6 +155,74 @@ func TestAttributeNilValue(t *testing.T) {
...
+
+	models := s.GetModels(func(m datalayer.Model) bool {
+		v, ok := m.GetAttributes().Get("vendor")
+		return ok && v.(testValue).Value == 1
+	})
+	if len(models) != 2 {
+		t.Errorf("expected 2 models matching predicate, got %d", len(models))
+	}
+
+	names := make(map[string]bool)
```
nirrozenbaum, line 196: nit: you could use k8s sets and use `Insert` and `Has` functions (instead of using map for set purposes)
not a blocker

nirrozenbaum:
/lgtm
/approve

@Jwrede Thanks! the PR looks good. I left few non blocking nits if you want to handle in a follow up

Jwrede:
Thanks for the review!

On the Model convenience methods: I would lean against that. The Model interface is currently 2 methods, and adding Put/Get/Delete would more than double its surface while duplicating the AttributeMap API. The chained `GetAttributes().Put(...)` pattern keeps Model minimal and follows standard Go composition. Happy to discuss further if you see it differently.

On the k8s sets nit: the repo does use `sets` in production code, but the existing `TestModels` test in the same file (line 98) uses the same `map[string]bool` pattern. I matched that for consistency. Happy to switch both if you prefer.

nirrozenbaum:
> Thanks for the review!
> 
> On the Model convenience methods: I would lean against that. The Model interface is currently 2 methods, and adding Put/Get/Delete would more than double its surface while duplicating the AttributeMap API. The chained `GetAttributes().Put(...)` pattern keeps Model minimal and follows standard Go composition. Happy to discuss further if you see it differently.

sure, that's fine also to keep it as is. I was thinking that from a developer perspective working with `model.SetAttribute(...)`, is nicer UX than the current way (AttributesMap might be even inline). but that's not a big issue to keep as is. 

> On the k8s sets nit: the repo does use `sets` in production code, but the existing `TestModels` test in the same file (line 98) uses the same `map[string]bool` pattern. I matched that for consistency. Happy to switch both if you prefer.

that's also just a nitpick. not a big deal to keep as is. using Sets is nice cause it's self documented (the intention in the map is set, not really any mapping here). if you want to change both places in a follow up that would be nice, but not a must. up to you :).

PR #152, opened 2026-06-02
Use profile picker

shmuelk:
**What type of PR is this?**
/kind feature

This PR causes the IPP to use the profile picker specified in the configuration text to select the profile containing the plugins to be run

?

```release-note
NONE
```

### DIFF
--- a/pkg/handlers/request.go
+++ b/pkg/handlers/request.go
@@ -67,7 +67,12 @@ func (s *Server) HandleRequestBody(ctx context.Context, reqCtx *RequestContext, 
 		return nil, errcommon.Error{Code: errcommon.BadRequest, Msg: fmt.Sprintf("failed to parse request body: %v", err)}
 	}
 
-	if err := s.runRequestPlugins(ctx, reqCtx.CycleState, reqCtx.Request); err != nil {
+	var err error
+	reqCtx.Profile, err = s.profilePicker.Pick(ctx, reqCtx.CycleState, reqCtx.Request, s.profiles)
+	if err != nil {
+		return nil, errcommon.Error{Code: errcommon.Internal, Msg: fmt.Sprintf("failed to pick a profile: %v", err)}
+	}
+	if err := s.runRequestPlugins(ctx, reqCtx.CycleState, reqCtx.Request, reqCtx.Profile); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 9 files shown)


pkg/handlers/response.go
```diff
@@ -131,7 +131,7 @@ func (s *Server) HandleResponseTrailers(trailers *eppb.HttpTrailers) ([]*eppb.Pr
 }
 
 // runResponsePlugins executes response plugins in the order they were registered.
-func (s *Server) runResponsePlugins(ctx context.Context, cycleState *plugin.CycleState, response *requesthandling.InferenceResponse) error {
+func (s *Server) runResponsePlugins(ctx context.Context, cycleState *plugin.CycleState, response *requesthandling.InferenceResponse, profile *requesthandling.Profile) error {
```
nirrozenbaum: nit: can we pass in only response plugins?

pkg/handlers/server.go
```diff
@@ -47,25 +47,26 @@ const (
...
+	profilePicker requesthandling.ProfilePicker
+	profiles      map[string]*requesthandling.Profile
 }
 
 // RequestContext stores context information during the lifetime of an HTTP request.
 type RequestContext struct {
 	RequestReceivedTimestamp    time.Time
 	ResponseFirstChunkTimestamp time.Time
 	ResponseCompleteTimestamp   time.Time
+	Profile                     *requesthandling.Profile
```
nirrozenbaum, line 69: nice

pkg/handlers/request.go
```diff
@@ -109,21 +114,21 @@ func (s *Server) HandleRequestBody(ctx context.Context, reqCtx *RequestContext,
 }
 
 // runRequestPlugins executes request plugins in the order they were registered.
-func (s *Server) runRequestPlugins(ctx context.Context, cycleState *plugin.CycleState, request *requesthandling.InferenceRequest) error {
+func (s *Server) runRequestPlugins(ctx context.Context, cycleState *plugin.CycleState, request *requesthandling.InferenceRequest,
+	profile *requesthandling.Profile) error {
```
nirrozenbaum: nit: can we pass in only request plugins and not the whole profile?

nirrozenbaum:
@shmuelk nice! the PR looks good.
I left two minor nits. 
also, would it be possible to add the pre processing plugins + post processing plugins and their execution (same as we have with profile picker).
then we should be able to configure them (e.g., always do the body to header plugin). 

/lgtm
/approve
/hold
holding in case you want to fix the nits. not a must, up to you.

PR #154, opened 2026-06-04
Refactor data layer supporting a unified mechanism for events, collectors and data sources

ronenkat:
**What type of PR is this?**
/kind feature

Refactor the data layer sources.
Run the data sources as a singleton supporting both extractors, collectors and data sources.

**Goal:** Replace two separately-configured plugins (notification-source, polling-source) with a single built-in Processor component that is always present and accessible to all plugins via the Handle.

What moved / merged
plugins/datalayer/pollingsource/plugin.go -->	interface/datalayer/datasource/processor.go
plugins/datalayer/notificationsource/plugin.go -->	merged into processor.go, file deleted

The Processor struct combines both: it runs one goroutine per Collector (polling) and one event-loop goroutine that dispatches Events to all Extractors (notification).

**Interface changes**
`datalayer/types.go` (renamed from datastore.go): added Event, EventType, and EventNotifier here to avoid an import cycle (plugin → datasource → plugin).
`datasource/types.go`: aliases those types from datalayer; removed NotificationSource and PollingSource interfaces; added DatalayerProcessor; added CollectorFrequency() to Collector.
`plugin/handle.go`: added EventNotifier() datalayer.EventNotifier to the Handle interface so any plugin can fire events into the processor without knowing about its lifecycle.

**Config changes**
```
API type: DataSources + PollingSources → single Datalayer []PluginRef
config.Config: DataSources/PollingSources → DatalayerSources []plugin.Plugin
YAML config example: datalayer: [{pluginRef: ...}] — one list for both collectors and extractors.

```
**Runtime wiring (runner.go)**
```
Creates datasource.NewProcessor() before building the Handle.
Passes the processor as the EventNotifier into NewHandle(...) so plugins can call handle.EventNotifier().Notify(e).
After config load, iterates DatalayerSources and registers each as a Collector, Extractor, or both.
Calls processor.Start(ctx) / defer processor.Stop() — no more separate start/stop loops.
```

```release-note
NONE
```


pkg/datalayer/processor.go
```diff
@@ -0,0 +1,224 @@
+/*
```
Mohammad-nassar10, line 1: The `interface` folder is conventionally used for interfaces and types definitions. The previous `NotificationSource` and `PollingSource` were in `pkg/framework/plugins/datalayer`. Is this `processor` a plugin as the `notificationSource`? Should we move it `pkg/framework/plugins/datalayer`?

ronenkat: The processor is not a plugin, but part of the datalayer.datasource package

pkg/config/loader/configloader.go
```diff
@@ -74,57 +73,32 @@ func LoadConfiguration(configBytes []byte, handle plugin.Handle, logger logr.Log
...
 
-func buildNotificationSources(refs []configapi.PluginRef, handle plugin.Handle) ([]datasource.NotificationSource, error) {
-	sources := make([]datasource.NotificationSource, 0, len(refs))
+func buildDatalayerSources(refs []configapi.PluginRef, handle plugin.Handle) ([]plugin.Plugin, error) {
+	sources := make([]plugin.Plugin, 0, len(refs))
 	for _, ref := range refs {
 		p := handle.Plugin(ref.PluginRef)
 		if p == nil {
 			return nil, fmt.Errorf("there is no plugin named %s", ref.PluginRef)
 		}
```
Mohammad-nassar10, line 137: in the Notification/Polling Source there was a check if the plugin is a desired type. This will silently skip non-matching plugins. Let's validate that it is a Collector or an Extractor.

pkg/framework/interface/datalayer/datasource/processor.go
```diff
@@ -0,0 +1,224 @@
...
+	logger := log.FromContext(ctx).WithName("processor")
+
+	for {
+		select {
+		case <-ctx.Done():
+			close(p.notifyDone)
+			return
+		case e := <-p.ch:
+			p.mu.Lock()
+			extractors := make([]Extractor, len(p.extractors))
```
Mohammad-nassar10: Why we need to update the extractor slice on every event? Is it valid to add extractors after the start? The new `Processor.RegisterExtractor` says "Safe to call before Start" (As we had in NotificationSource).

apix/config/v1alpha1/payloadprocessorconfig_types.go
```diff
@@ -64,12 +64,9 @@ type PayloadProcessorConfig struct {
...
 	// +optional
-	// NotificationSources is an optional list of references to notification-source plugins to start.
-	NotificationSources []PluginRef `json:"notificationSources,omitempty"`
-
-	// +optional
-	// PollingSources is an optional list of references to polling-source plugins to start.
-	PollingSources []PluginRef `json:"pollingSources,omitempty"`
+	// Datalayer is an optional list of references to Collector and Extractor plugins
+	// registered with the built-in datalayer Processor.
+	Datalayer []PluginRef `json:"datalayer,omitempty"`
```
Mohammad-nassar10: A flat `[]PluginRef` mixes collectors and extractors with no distinction in the schema. What about splitting into two named fields?

pkg/config/loader/configloader.go
ronenkat: Thanks. moved the registration logic to here.

pkg/framework/interface/datalayer/datasource/processor.go
ronenkat: The snapshot allows adding extractor during running. This follows the same pattern as enabling collectors to be added

apix/config/v1alpha1/payloadprocessorconfig_types.go
ronenkat: What would be the value?

pkg/datalayer/processor.go
shmuelk: pkg/framework/interface/datalayer/datasource should contain types and interfaces, not implementations.

apix/config/v1alpha1/payloadprocessorconfig_types.go
Mohammad-nassar10: Collectors []PluginRef `json:"collectors,omitempty"`
Extractors []PluginRef `json:"extractors,omitempty"`
?

pkg/framework/interface/datalayer/datasource/processor.go
shmuelk: I agree that in general once started we shouldn't add extra Extractors.

pkg/framework/interface/plugin/handle.go
```diff
@@ -105,11 +112,16 @@ func (h *ippHandlePlugins) GetAllPluginsWithNames() map[string]Plugin {
 	return h.plugins
 }
 
-func NewHandle(ctx context.Context, mgr ctrl.Manager, ds datalayer.Datastore) Handle {
+func NewHandle(ctx context.Context, mgr ctrl.Manager, ds datalayer.Datastore, notifier ...datalayer.EventNotifier) Handle {
+	var n datalayer.EventNotifier
+	if len(notifier) > 0 {
+		n = notifier[0]
+	}
```
shmuelk: Simply pass in nil, where you don't need a notifier (I assume that in tests...)

apix/config/v1alpha1/payloadprocessorconfig_types.go
ronenkat: user value?
The plugin are sorted in the processor to types anyway.

pkg/framework/interface/plugin/handle.go
ronenkat: Thanks. updated.

apix/config/v1alpha1/payloadprocessorconfig_types.go
Mohammad-nassar10: yes, more readable.
But it is not blocking

pkg/datalayer/processor.go
ronenkat: moved to pkg/datalyer.
Merged datalayer.datasource.types.go into datalayer.types.go

pkg/framework/interface/datalayer/datasource/processor.go
ronenkat: updated.

Mohammad-nassar10 (approved):

cmd/runner/runner.go
```diff
@@ -255,15 +239,13 @@ func (r *Runner) loadConfiguration(ctx context.Context, opts *runserver.Options,
 	// Register factories for all known in-tree plugins
 	r.registerInTreePlugins()
 
-	theConfig, err := loader.LoadConfiguration(configBytes, handle, logger)
+	theConfig, err := loader.LoadConfiguration(configBytes, handle, r.processor, logger)
```
shmuelk, line 242: Why are you passing in the processor twice? Once in the Handle and once as a direct parameter?

pkg/config/loader/configloader.go
```diff
@@ -74,59 +74,66 @@ func LoadConfiguration(configBytes []byte, handle plugin.Handle, logger logr.Log
...
-		PollingSources:      pollingSources,
+		ProfilePicker: profilePicker,
+		Profiles:      profiles,
 	}, nil
 }
 
-func buildNotificationSources(refs []configapi.PluginRef, handle plugin.Handle) ([]datasource.NotificationSource, error) {
-	sources := make([]datasource.NotificationSource, 0, len(refs))
-	for _, ref := range refs {
+func buildDatalayerSources(cfg *configapi.DatalayerConfig, handle plugin.Handle, processor datasource.DatalayerProcessor) error {
```
shmuelk, line 107: Same comment here. Why is the processor passed in twice?

pkg/config/loader/configloader.go
```diff
@@ -74,59 +74,66 @@ func LoadConfiguration(configBytes []byte, handle plugin.Handle, logger logr.Log
...
-			return nil, fmt.Errorf("there is no plugin named %s", ref.PluginRef)
+			return fmt.Errorf("there is no plugin named %s", ref.PluginRef)
 		}
-		src, ok := p.(datasource.PollingSource)
+		d, ok := p.(datasource.DataSource)
 		if !ok {
-			return nil, fmt.Errorf("the plugin named %s is not a PollingSource", ref.PluginRef)
+			return fmt.Errorf("plugin %q is not a DataSource", ref.PluginRef)
+		}
+		if processor != nil {
```
shmuelk: Why should it ever be nil? Tests that include the Datasource/DataLayer section should include a processor of some kind.

cmd/runner/runner.go
ronenkat: The Handle include only EventNotifier() interface, not the registration interface (collectors, datasources, extractors), which is included in the processor interface.
The Handle is used to resolve the plugins from the configuration (as string).
They have different uses.

pkg/config/loader/configloader.go
ronenkat: Here is the root cause, which is to register the data layer plugins with the processor.

pkg/config/loader/configloader.go
ronenkat: updated.

cmd/runner/runner.go
ronenkat: As the handle is passed to the plugins, it only includes the notify interface

PR #155, opened 2026-06-04
Load Pre-Processors and P post processor config

shmuelk:
**What type of PR is this?**
/kind feature

This PR loads the configuration with the pre-processor and post-processor plugins from the textual configuration

```release-note
NONE
```

### DIFF
--- a/pkg/config/loader/configloader_test.go
+++ b/pkg/config/loader/configloader_test.go
@@ -48,6 +48,8 @@ const (
 	testFilterType       = "test-filter"
 	testPickerType       = "test-picker"
 	testPluginType       = "test-plugin"
+	testPostProcessor    = "test-post-processor"
+	testPreProcessor     = "test-pre-processor"
 	testProfilePicker    = "test-profile-picker"
 	testRequestProcType  = "test-request-processor"
 	testResponseProcType = "test-response-processor"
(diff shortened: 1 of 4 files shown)


pkg/config/loader/configloader_test.go
```diff
@@ -326,6 +328,88 @@ func TestBuildProfiles(t *testing.T) {
...
+			handle := plugin.NewHandle(context.Background(), nil, nil)
+			err = instantiatePlugins(rawConfig.Plugins, handle)
+			require.NoError(t, err, "setup: instantiatePlugins failed")
+
+			preProcessors, preErr := buildPreProcessors(rawConfig.PreProcessing, handle)
+			postProcessors, postErr := buildPostProcessors(rawConfig.PostProcessing, handle)
+
+			if tc.wantErr {
+				if preErr == nil && postErr == nil {
+					t.Logf("either buildPreProcessors or buildPostProcessors was suppose to fail")
```
nirrozenbaum: nit: t.Logf only logs — it does not fail the test. If both builders unexpectedly succeed, the test silently passes.

nirrozenbaum:
/lgtm
/approve
/hold

I left one minor comment about a bug in the test (passes silently although it should fail).
if you want to fix in a follow up feel free to unhold.

nirrozenbaum:
now it looks like bug is fixed.
/lgtm
/approve

PR #159, opened 2026-06-12
fix(observability): extract upstream traceparent, re-parent server span, and inject traceparent on egress

### LINKED ISSUE #157: [observability] IPP: extract upstream traceparent at ext_proc ingress, re-parent span, and inject traceparent on egress

- In the `RequestHeaders` branch of `Process()`, extract upstream trace context from the Envoy header map using `propagator.Extract`.
- Move the server span start to after extraction, or re-parent it to the extracted context (avoid the orphan root).

**Why is this needed**:

IPP runs as an Envoy `ext_proc` filter (`config/charts/payload-processor/templates/istio.yaml`), sequentially on the same Envoy HTTP filter chain as EPP. Like EPP, IPP starts its server span at the top of `Process()` before the request headers are received, so it never extracts the upstream `traceparent` and starts a new root trace per request. In addition, IPP does not inject `traceparent` on egress: it already has a HeaderMutation pipeline but trace context is not written into it, so a downstream `ext_proc` filter in the chain may not see the updated `traceparent`.

gyliu513:
**What type of PR is this?**

/kind feature

IPP runs as an Envoy `ext_proc` filter, sequentially on the same Envoy HTTP filter chain as EPP. Like EPP (fixed in llm-d/llm-d-router#1514), IPP started its server span at the top of `Process()` before the request headers were received, so it never extracted the upstream `traceparent` and started a new root trace per request.

In addition, unlike EPP, IPP did not inject `traceparent` on egress: it already has a HeaderMutation pipeline, but trace context was never written into it, so a downstream `ext_proc` filter in the chain could not see IPP's span as its parent.

Fixes #157 

```release-note
IPP now extracts the upstream trace context from request headers and re-parents its server span to it, and injects `traceparent` into the egress header mutation so downstream ext_proc filters join the same trace.
```


pkg/handlers/request.go
```diff
@@ -46,15 +48,32 @@ func (s *Server) HandleRequestHeaders(ctx context.Context, reqCtx *RequestContex
 		}
 	}
 
+	// Inject the active trace context into the egress headers so the next
+	// filter in the chain is parented to this span rather than the upstream one.
+	traceCarrier := propagation.MapCarrier{}
```
shaneutt, line 57: Do we want to normalize headers from the tracecarrier as LC, as a "just in case"? 🤔

gyliu513: Good call, it's a no-op for the default W3C propagator (traceparent/tracestate are already lowercase), but SetHeader writes the key verbatim into the Envoy header mutation, and over HTTP/2 a non-W3C/custom propagator could emit mixed-case keys.

Lowercasing is cheap defensive insurance, so I've normalized the carrier keys with strings.ToLower.

pkg/handlers/request.go
```diff
@@ -46,15 +49,34 @@ func (s *Server) HandleRequestHeaders(ctx context.Context, reqCtx *RequestContex
...
+	otel.GetTextMapPropagator().Inject(ctx, traceCarrier)
+	for key, value := range traceCarrier {
+		// Normalize keys to lowercase; HTTP/2 (used by ext_proc) requires
+		// lowercase header names, and a non-W3C propagator may emit mixed case.
+		reqCtx.Request.SetHeader(strings.ToLower(key), value)
```
shmuelk, line 62: This will cause duplicate keys, differenciated only by their case.

gyliu513: Thanks @shmuelk , I will normalize incoming header keys to lowercase at ingestion

gyliu513:
@shmuelk @shaneutt @nirrozenbaum can you help review?

gyliu513:
@shmuelk @shaneutt @nirrozenbaum can you help review? I think all comments are now addressed.

gyliu513:
CI is being fixed at https://github.com/llm-d/llm-d-inference-payload-processor/pull/261

PR #165, opened 2026-06-13
feat(observability): add fine-grained OTel spans for the plugin pipeline and model selector

### LINKED ISSUE #163: [observability] Add fine-grained OpenTelemetry spans for the plugin pipeline and model selector, aligned with llm-d-router#1483
## Description

Add fine-grained OpenTelemetry spans for IPP's request/response plugin pipeline and the model-selector pipeline (filters → scorers → picker), so operators can see *which plugins ran, in what order, and how long each took* inside a single request trace — not only the top-level `gateway.request` span.

This aligns IPP with llm-d/llm-d-router#1483, which adds the equivalent scorer-chain spans to the EPP scheduling pipeline. IPP and EPP run on the same Envoy filter chain and appear in the same traces (see #157), so a sampled request should show comparable pipeline detail on both hops.

## Current behavior

The entire request lifecycle produces exactly one span: `gateway.request`, started at the top of `Process()` in `pkg/handlers/server.go`. Everything below it is opaque in trace UIs.

The instrumentation points already exist — each pipeline stage records a `ipp_plugin_duration_seconds` histogram via `RecordPluginProcessingLatency`:

- request pre-processors loop in `pkg/handlers/request.go`
- response post-processors loop in `pkg/handlers/response.go`
- model-selector filters, scorers, and picker in `pkg/modelselector/model_selector_pipeline.go`
- model-selector end-to-end in `pkg/modelselector/modelselector.go`

Metrics give aggregate latency per plugin, but they cannot tie a *single request* through the pipeline: when one request is slow or a model selection is wrong, there is no way to see which plugin dominated or which filters/scorers actually ran for that request.

## Proposed change

Mirror the structure agreed in llm-d-router#1483:

1. **Parent spans per stage**: a child span under `gateway.request` for each pipeline stage, e.g. `llm_d.ipp.request_plugins`, `llm_d.ipp.model_selector`, `llm_d.ipp.response_plugins`.
2. **Uniform per-plugin sub-spans via a shared wrapper**: wrap plugin invocation once at the existing `RecordPluginProcessingLatency` call sites (or a decorator), emitting spans like `llm_d.ipp.plugin.<type>` with bounded-cardinality attributes: `llm_d.plugin.extension_point`, `llm_d.plugin.type`, `llm_d.plugin.name`, and for scorers `llm_d.scorer.candidate_count` / score aggregates. No per-model or per-request-content attributes beyond what #160's conventions allow.
3. **Model-selector outcome attributes**: on the `llm_d.ipp.model_selector` span, record the incoming requested model and the selected model — this is IPP's core decision and the single most useful thing to see in a trace.
4. **Naming**: follow the service/scope conventions being standardized in #160 (aligned with llm-d/llm-d-router#1519), using `llm_d.ipp.*` to parallel EPP's `llm_d.epp.*`.

## Acceptance criteria

- With tracing enabled and a sampled request, Jaeger (or equivalent) shows: `gateway.request` → `llm_d.ipp.request_plugins` / `llm_d.ipp.model_selector` (→ filter/scorer/picker spans) / `llm_d.ipp.response_plugins`.
- Existing Prometheus metrics (`ipp_plugin_duration_seconds`, `ipp_model_selector_*`) remain unchanged; spans complement metrics, they do not replace them.

## Why is this needed

- **Operations**: when routing is wrong or slow, the questions are per-request — *which scorer dominated latency? what did the picker choose and why?* Aggregate histograms cannot answer them; spans can.
- **Low cost**: all instrumentation points already exist as metric call sites; a shared wrapper adds spans without N copy-paste `tracer.Start` calls and keeps cardinality bounded (plugin names are operator-configured).
- **Cross-component consistency**: once llm-d/llm-d-router#1483 lands, traces will show fine-grained detail on the EPP hop but a single opaque span on the IPP hop of the same request.

/cc @nirrozenbaum

gyliu513:
**What type of PR is this?**

/kind feature

Adds fine-grained OpenTelemetry spans for IPP's request/response plugin pipeline and the model-selector pipeline (filters → scorers → picker), so a single sampled request shows *which plugins ran, in what order, and how long each took*, instead of one opaque `gateway.request` span.

Until now the entire request lifecycle produced exactly one span.

Fixes #163 

```release-note
Adds fine-grained OpenTelemetry spans for IPP's request/response plugin pipeline and the model-selector pipeline
```


pkg/handlers/request.go
```diff
@@ -131,17 +135,34 @@ func (s *Server) runRequestPlugins(ctx context.Context, cycleState *plugin.Cycle
 	verboseLogger := logger.V(logutil.VERBOSE)
 	verboseEnabled := verboseLogger.Enabled()
 
+	// Stage span grouping the per-plugin spans under gateway.request.
+	tracer := tracing.Tracer(handlersTracerScope)
+	ctx, stageSpan := tracer.Start(ctx, "request_plugins", trace.WithSpanKind(trace.SpanKindInternal))
+	defer stageSpan.End()
+
 	for _, reqPlugin := range reqPlugins {
+		name := reqPlugin.TypedName()
```
nirrozenbaum: nit: maybe better to name this `typedName`?
it reads a bit strange `name.Type` and `name.Name`.

```suggestion
		typedName := reqPlugin.TypedName()
```

pkg/handlers/request.go
```diff
@@ -131,17 +135,34 @@ func (s *Server) runRequestPlugins(ctx context.Context, cycleState *plugin.Cycle
...
-			verboseLogger.Info("Executing request plugin", "plugin", reqPlugin.TypedName())
+			verboseLogger.Info("Executing request plugin", "plugin", name)
 		}
+		pluginCtx, span := tracer.Start(ctx, "plugin."+name.Type,
+			trace.WithSpanKind(trace.SpanKindInternal),
+			trace.WithAttributes(
+				attribute.String("llm_d.plugin.extension_point", requestPluginExtensionPoint),
+				attribute.String("llm_d.plugin.type", name.Type),
+				attribute.String("llm_d.plugin.name", name.Name),
+			))
```
nirrozenbaum: I'm not a tracing expert so forgive me if the question is obvious.. 
why do we specify type twice? 
once in plugin.name.Type and another one in attribute?

pkg/handlers/response.go
```diff
@@ -151,18 +155,34 @@ func (s *Server) runResponsePlugins(ctx context.Context, cycleState *plugin.Cycl
...
 	verboseEnabled := verboseLogger.Enabled()
 
-	var err error
+	// Stage span grouping the per-plugin spans under gateway.request.
+	tracer := tracing.Tracer(handlersTracerScope)
+	ctx, stageSpan := tracer.Start(ctx, "response_plugins", trace.WithSpanKind(trace.SpanKindInternal))
+	defer stageSpan.End()
+
 	for _, respPlugin := range respPlugins {
+		name := respPlugin.TypedName()
```
nirrozenbaum: ditto, typedName

pkg/modelselector/model_selector_pipeline.go
```diff
@@ -172,19 +180,31 @@ func (p *ModelSelectorPipeline) runFilterPlugins(ctx context.Context, request *r
 		debugLogger.Info("Before running filter plugins", "models", len(filteredModels))
 	}
 
+	tracer := tracing.Tracer(modelSelectorTracerScope)
 	for _, filter := range p.filters {
+		name := filter.TypedName()
```
nirrozenbaum: ditto

pkg/modelselector/model_selector_pipeline.go
```diff
@@ -252,14 +284,27 @@ func (p *ModelSelectorPipeline) runPickerPlugin(ctx context.Context, cycleState
...
+			attribute.Int("llm_d.picker.candidate_count", len(scoredModels)),
+		))
 	before := time.Now()
-	result := p.picker.Pick(ctx, cycleState, scoredModels)
-	metrics.RecordPluginProcessingLatency(pickerExtensionPoint, p.picker.TypedName().Type, p.picker.TypedName().Name, time.Since(before))
+	result := p.picker.Pick(spanCtx, cycleState, scoredModels)
+	metrics.RecordPluginProcessingLatency(pickerExtensionPoint, name.Type, name.Name, time.Since(before))
+	if result != nil && result.TargetModel != nil {
+		span.SetAttributes(attribute.String("llm_d.picker.selected_model", result.TargetModel.GetName()))
+	}
```
nirrozenbaum: Picker is guaranteed to select a target model. we can remove the conditional.

```suggestion
	span.SetAttributes(attribute.String("llm_d.picker.selected_model", result.TargetModel.GetName()))
```

pkg/modelselector/modelselector.go
```diff
@@ -53,10 +57,26 @@ func (s *ModelSelector) Select(ctx context.Context, request *requesthandling.Inf
...
+
 	selectStart := time.Now()
 	defer func() {
 		metrics.RecordModelSelectorE2ELatency(time.Since(selectStart))
 		metrics.RecordModelSelectorAttempt(err)
+		if err != nil {
+			span.RecordError(err)
+			span.SetStatus(codes.Error, err.Error())
+		} else if result != nil && result.TargetModel != nil {
+			span.SetAttributes(attribute.String("llm_d.model_selector.selected_model", result.TargetModel.GetName()))
```
nirrozenbaum: this would fit better before the log line `Model selection completed`. without the conditionals.

nirrozenbaum (commented):
@gyliu513 thanks, overall looks good. 
left few minor comments.

pkg/handlers/request.go
gyliu513: Done, renamed to typedName here and in the response/filter/scorer/picker loops for consistency. Thanks.

pkg/handlers/request.go
gyliu513: Good question, it's intentional.

The span name (plugin.type)is the human-readable label shown in trace UIs; it's meant to be low-cardinality and isn't reliably queryable across tracing backends. 

The `llm_d.plugin.type` attribute is the structured, indexable dimension you filter and aggregate on (e.g. "p99 latency grouped by plugin type"). Per OTel conventions, the span name is for display and attributes carry the queryable data, so keeping both is by design.

pkg/handlers/response.go
gyliu513: Done

pkg/modelselector/model_selector_pipeline.go
gyliu513: Done

pkg/modelselector/modelselector.go
gyliu513: Done

pkg/modelselector/model_selector_pipeline.go
gyliu513: Done

gyliu513 (commented):
Thanks @nirrozenbaum

pkg/handlers/request.go
```diff
@@ -131,17 +135,34 @@ func (s *Server) runRequestPlugins(ctx context.Context, cycleState *plugin.Cycle
...
+			span.RecordError(err)
+			span.SetStatus(codes.Error, err.Error())
+			span.End()
+			logger.Error(err, "Failed to execute request plugin", "plugin", typedName)
 			return err
 		}
+		span.End()
 	}
 
 	return nil
```
shaneutt, line 197: Seems that we may want to add more test coverage for these changes:

We need a test that does this:

1. Set up a `tracetest.SpanRecorder` (as done in `telemetry_test.go`).
2. Run `runRequestPlugins` with one or more fake plugins.
3. Assert that a `request_plugins` stage span is created with child `plugin.*` spans.
4. Assert that a failing plugin produces a span with `codes.Error` status.

If I'm not mistaken, I don't believe we have that currently which would leave an opportunity for regressions.

pkg/handlers/server.go
shaneutt: Should we now use the new `tracing.Tracer` helper?

shaneutt: A new constant for this was added above but we're not using it here yet.

pkg/modelselector/model_selector_pipeline.go
```diff
@@ -252,14 +284,25 @@ func (p *ModelSelectorPipeline) runPickerPlugin(ctx context.Context, cycleState
...
+		trace.WithAttributes(
+			attribute.String("llm_d.plugin.extension_point", pickerExtensionPoint),
+			attribute.String("llm_d.plugin.type", typedName.Type),
+			attribute.String("llm_d.plugin.name", typedName.Name),
+			attribute.Int("llm_d.picker.candidate_count", len(scoredModels)),
+		))
 	before := time.Now()
-	result := p.picker.Pick(ctx, cycleState, scoredModels)
-	metrics.RecordPluginProcessingLatency(pickerExtensionPoint, p.picker.TypedName().Type, p.picker.TypedName().Name, time.Since(before))
+	result := p.picker.Pick(spanCtx, cycleState, scoredModels)
```
shaneutt, line 302: It appears that this call may be able to return a `nil` value for `TargetModel`?

If it does we will panic on `GetName()` below.

pkg/modelselector/model_selector_pipeline.go
```diff
@@ -252,14 +284,25 @@ func (p *ModelSelectorPipeline) runPickerPlugin(ctx context.Context, cycleState
...
+	metrics.RecordPluginProcessingLatency(pickerExtensionPoint, typedName.Type, typedName.Name, time.Since(before))
+	span.SetAttributes(attribute.String("llm_d.picker.selected_model", result.TargetModel.GetName()))
+	span.End()
 	if debugEnabled {
-		debugLogger.Info("Completed running picker plugin", "plugin", p.picker.TypedName(), "result", result)
+		debugLogger.Info("Completed running picker plugin", "plugin", typedName, "result", result)
 	}
 
 	return result
```
shaneutt, line 310: I think we're in a situation with this one, as before, where test coverage needs expansion.

pkg/modelselector/modelselector.go
```diff
@@ -75,6 +93,7 @@ func (s *ModelSelector) Select(ctx context.Context, request *requesthandling.Inf
 		return nil, err
 	}
 
+	span.SetAttributes(attribute.String("llm_d.model_selector.selected_model", result.TargetModel.GetName()))
```
shaneutt, line 96: Similar to above: no `nil` guards

pkg/handlers/request.go
gyliu513: I'll add a `runRequestPlugins` test using a `tracetest.SpanRecorder` (mirroring `telemetry_test.go`) with fake plugins that asserts the `request_plugins` stage span, nested `plugin.*` child spans, and a `codes.Error` status when a plugin fails.

pkg/handlers/server.go
gyliu513: Yes, I'll fix it. It was merged after I created this PR.

gyliu513: Yes, I'll fix it. It was merged after I created this PR.

pkg/modelselector/model_selector_pipeline.go
gyliu513: Yes, will add a test case

pkg/modelselector/modelselector.go
gyliu513: This one is already safe — lines 91-94 guard immediately

pkg/modelselector/model_selector_pipeline.go
gyliu513: Hi @shaneutt , @nirrozenbaum post a comment [here](https://github.com/llm-d/llm-d-inference-payload-processor/pull/165#discussion_r3409523088) and I agree that the Picker is guaranteed to select a target model, do we still need the guard?

pkg/common/observability/tracing/telemetry.go
```diff
@@ -24,17 +24,40 @@ import (
...
 	semconv "go.opentelemetry.io/otel/semconv/v1.37.0"
+	"go.opentelemetry.io/otel/trace"
 
 	"github.com/llm-d/llm-d-inference-payload-processor/pkg/common/observability/logging"
 	"github.com/llm-d/llm-d-inference-payload-processor/version"
 )
 
+// instrumentationName is the default OTel instrumentation scope used when no
+// explicit scope is supplied to Tracer.
+const instrumentationName = "llm-d-inference-payload-processor"
```
shmuelk, line 42: I recommend shortning the name. How about:


```suggestion
const instrumentationName = "llm-d-ipp"
```

pkg/common/observability/tracing/telemetry.go
```diff
@@ -24,17 +24,40 @@ import (
...
+
+// Tracer returns a tracer for the given instrumentation scope, defaulting to
+// "llm-d-inference-payload-processor". The build version and commit SHA are
```
shmuelk, line 45:
```suggestion
// "llm-d-ipp". The build version and commit SHA are
```

pkg/handlers/server.go
```diff
@@ -46,6 +44,10 @@ const (
 
 	requestPluginExtensionPoint  = "request"
 	responsePluginExtensionPoint = "response"
+
+	// handlersTracerScope is the OTel instrumentation scope for spans emitted by
+	// the request/response handlers, following the package-path naming convention.
+	handlersTracerScope = "llm-d-inference-payload-processor/pkg/handlers"
```
gyliu513: @shmuelk with your above comment, do we need to update here as well to 

```suggestion
	handlersTracerScope = "llm-d-ipp/pkg/handlers"
```

/cc @nirrozenbaum @shaneutt

pkg/modelselector/model_selector_pipeline.go
```diff
@@ -40,6 +43,11 @@ const (
 	filterExtensionPoint = "ModelSelectorFilter"
 	scorerExtensionPoint = "ModelSelectorScorer"
 	pickerExtensionPoint = "ModelSelectorPicker"
+
+	// modelSelectorTracerScope is the OTel instrumentation scope for spans
+	// emitted by the model-selector pipeline, following the package-path
+	// naming convention.
+	modelSelectorTracerScope = "llm-d-inference-payload-processor/pkg/modelselector"
```
gyliu513: @shmuelk ditto here, how about


```suggestion
modelSelectorTracerScope = "llm-d-ipp/pkg/modelselector"
```

gyliu513:
@shmuelk @shaneutt @nirrozenbaum can you help review?

gyliu513:
@shmuelk @shaneutt @nirrozenbaum can you help review? I think all comments are addressed.

pkg/handlers/server.go
shmuelk: yes

pkg/modelselector/model_selector_pipeline.go
shmuelk: yes, again

pkg/handlers/server.go
gyliu513: Done

pkg/modelselector/model_selector_pipeline.go
gyliu513: Done

shmuelk:
A similar PR is being discussed in the ll-d-router repo.Some, including myself think this very low level tracing is too much.

All of my requested changes have been made.

From my perspective this PR can be merged if the functionality is desired and not deemed to be too much.

gyliu513:
@shmuelk I was talking with @elevran at https://github.com/llm-d/llm-d-router/pull/1834, but I think we should be good to merge this.

gyliu513:
Hey @nirrozenbaum , I think this can be merged, can you help check?

pkg/handlers/request.go
```diff
@@ -160,17 +164,34 @@ func (s *Server) runRequestPlugins(ctx context.Context, cycleState *plugin.Cycle
...
 		before := time.Now()
-		err := reqPlugin.ProcessRequest(ctx, cycleState, request)
-		metrics.RecordPluginProcessingLatency(requestPluginExtensionPoint, reqPlugin.TypedName().Type, reqPlugin.TypedName().Name, time.Since(before))
+		err := reqPlugin.ProcessRequest(pluginCtx, cycleState, request)
+		metrics.RecordPluginProcessingLatency(requestPluginExtensionPoint, typedName.Type, typedName.Name, time.Since(before))
 		if err != nil {
-			logger.Error(err, "Failed to execute request plugin", "plugin", reqPlugin.TypedName())
+			span.RecordError(err)
+			span.SetStatus(codes.Error, err.Error())
+			span.End()
```
nirrozenbaum: nit: can we have 
```
defer span.End()
```
only once after we create it in L177 instead of having writing it twice?

pkg/handlers/response.go
```diff
@@ -286,18 +290,34 @@ func (s *Server) runResponsePlugins(ctx context.Context, cycleState *plugin.Cycl
...
 		before := time.Now()
-		err = respPlugin.ProcessResponse(ctx, cycleState, response)
-		metrics.RecordPluginProcessingLatency(responsePluginExtensionPoint, respPlugin.TypedName().Type, respPlugin.TypedName().Name, time.Since(before))
+		err := respPlugin.ProcessResponse(pluginCtx, cycleState, response)
+		metrics.RecordPluginProcessingLatency(responsePluginExtensionPoint, typedName.Type, typedName.Name, time.Since(before))
 		if err != nil {
-			logger.Error(err, "Failed to execute response plugin", "plugin", respPlugin.TypedName())
+			span.RecordError(err)
+			span.SetStatus(codes.Error, err.Error())
+			span.End()
```
nirrozenbaum: ditto

pkg/handlers/server.go
```diff
@@ -128,13 +131,7 @@ func (s *Server) Process(srv extProcPb.ExternalProcessor_ProcessServer) error {
 	// The server span is started when the request headers arrive, so it can be
 	// parented to the upstream trace context they carry instead of starting an
 	// orphan root trace.
-	tracer := otel.Tracer(
-		"llm-d-inference-payload-processor/pkg/handlers",
-		trace.WithInstrumentationVersion(version.BuildRef),
-		trace.WithInstrumentationAttributes(
-			attribute.String("commit-sha", version.CommitSHA),
```
nirrozenbaum: don't we want to include buildRef and commitSHA?
those sound like important attributes

pkg/modelselector/model_selector_pipeline.go
```diff
@@ -174,19 +182,31 @@ func (p *ModelSelectorPipeline) runFilterPlugins(ctx context.Context, request *r
...
 	}
 
+	tracer := tracing.Tracer(modelSelectorTracerScope)
 	for _, filter := range p.filters {
+		typedName := filter.TypedName()
 		if verboseEnabled {
-			verboseLogger.Info("Running filter plugin", "plugin", filter.TypedName())
+			verboseLogger.Info("Running filter plugin", "plugin", typedName)
 		}
+		spanCtx, span := tracer.Start(ctx, "plugin."+typedName.Type,
```
nirrozenbaum: what happens if there are multiple plugins from the same type? 
isn't it better to have `plugin. +typedName.String()' to include both type and name here?

pkg/modelselector/model_selector_pipeline.go
```diff
@@ -216,20 +236,32 @@ func (p *ModelSelectorPipeline) runScorerPlugins(ctx context.Context, request *r
...
 	}
 
+	tracer := tracing.Tracer(modelSelectorTracerScope)
 	for _, scorer := range p.scorers {
+		typedName := scorer.TypedName()
 		if verboseEnabled {
-			verboseLogger.Info("Running scorer plugin", "plugin", scorer.TypedName())
+			verboseLogger.Info("Running scorer plugin", "plugin", typedName)
 		}
+		spanCtx, span := tracer.Start(ctx, "plugin."+typedName.Type,
```
nirrozenbaum: ditto

pkg/modelselector/model_selector_pipeline.go
```diff
@@ -216,20 +236,32 @@ func (p *ModelSelectorPipeline) runScorerPlugins(ctx context.Context, request *r
...
+			trace.WithSpanKind(trace.SpanKindInternal),
+			trace.WithAttributes(
+				attribute.String("llm_d.plugin.extension_point", scorerExtensionPoint),
+				attribute.String("llm_d.plugin.type", typedName.Type),
+				attribute.String("llm_d.plugin.name", typedName.Name),
+				attribute.Int("llm_d.scorer.candidate_count", len(models)),
+				attribute.Float64("llm_d.scorer.weight", scorer.Weight()),
```
nirrozenbaum: I think this should come after the `Score` function and include the score as well. 
ideally one would like to see in the traces the scorer typed name, the score and and the weight.
score is missing cause it's too early.

nirrozenbaum (commented):
@gyliu513 thank you for the PR.
left some comments/questions

pkg/handlers/request.go
gyliu513: Good catch on the duplication, but a plain `defer span.End()` right after `tracer.Start()` here wouldn't do what we want, since the span is created inside the `for` loop body of this function (not a per-iteration helper), the defer would only fire when `runRequestPlugins` returns, not per plugin, so all the per-plugin spans would stay open until the loop finishes and get bogus end-times. `llm-d-router`'s `runScorer` (`scheduler_profile.go:288-289`) hits the same shape and solves it by extracting the per-plugin span+call into its own function so the defer scopes correctly per call. I can refactor `runRequestPlugins`/`runResponsePlugins` the same way if you'd like, happy to do that as a follow-up in this PR.

pkg/handlers/response.go
gyliu513: same as above reply

pkg/handlers/server.go
gyliu513: They're still there, just centralized, `tracing.Tracer()` (`telemetry.go:52-58`) attaches `WithInstrumentationVersion(version.BuildRef)` and the `commit-sha` attribute to every tracer it returns, so `Process`'s spans get them via the `tracing.Tracer(handlersTracerScope)` call a few lines up. Removed the duplicate inline `otel.Tracer(...)` call so it's set in one place instead of every call site.

pkg/modelselector/model_selector_pipeline.go
gyliu513: This mirrors the convention `llm-d-router` uses for its scorer spans (`scheduler_profile.go:288`), span name carries just the Type for low cardinality, and both `llm_d.plugin.type`/`llm_d.plugin.name` are separate attributes for the actual instance. If two plugin instances share a Type, their spans will look identical by name in the waterfall, but the attributes (visible on hover in Jaeger/Tempo/etc.) disambiguate them. Keeping it consistent with `llm-d-router`'s existing pattern rather than introducing a different naming scheme here, but open to revisiting org-wide if you feel disambiguation-by-name is worth the cardinality cost.

pkg/modelselector/model_selector_pipeline.go
gyliu513: Agreed, this is worth fixing. `llm-d-router` (`scheduler_profile.go:297-316`) and `llm-d-kv-cache` (`traced_scorer.go:60-82`) both do this: `span.SetAttributes(...)` right after `Score()` returns, recording aggregate `score.max`/`score.avg` (not per-model, to keep cardinality bounded). I'll follow the same pattern here — add `llm_d.scorer.score.max`/`llm_d.scorer.score.avg` computed from `scores` right before `span.End()`.

pkg/handlers/request.go
nirrozenbaum: yeah extracting to a helper function would make it nicer :)

pkg/modelselector/model_selector_pipeline.go
nirrozenbaum: I'm not a tracing expert, but I think it might be worth considering revisiting this point. 
looking on a waterfall of spans, it should be clear what each refers to without having to drill down.
e.g., if we see multiple spans like this there's no way to understand without drilling down:
https://github.com/llm-d/llm-d-router/blob/db17418ade6cd5a15c6bbb04c9d0f0089c37034f/pkg/epp/framework/plugins/scheduling/scorer/endpointattribute/endpointattribute.go#L32

pkg/modelselector/model_selector_pipeline.go
nirrozenbaum: apologize for picking to this point. 
I'm not sure max/avg gives us information that's really helpful in terms of observability. 
I was thinking on a more detailed option, something like creating a new span `scores` and then have a entry with two attributes - target and score (and I would recommend doing the same in llm-d-router). 
that could allow us to drill into a specific request and understand what was the score for each of the models/endpoints, the span one level up can specify the selected target. 
in case max-picker is used, then obviously it's the target with max score, but in case a different picker is used, it's not obvious which would get selected.

pkg/handlers/request.go
gyliu513: Fixed

pkg/modelselector/model_selector_pipeline.go
gyliu513: Fixed by switching the span name from `"plugin."+typedName.Type` to `"plugin."+typedName.String()` (`plugin.TypedName.String()` already renders `<name>/<type>`) in the filter and scorer loops here, and did the same in `runRequestPlugin`/`runResponsePlugin` in `pkg/handlers` for consistency. Now e.g. two same-`Type` scorers show up as distinct spans like `plugin.queue-depth/endpoint-attribute-scorer` and `plugin.kv-cache-util/endpoint-attribute-scorer` without drilling into attributes.

pkg/modelselector/model_selector_pipeline.go
gyliu513: Makes sense, dropped the max/avg attributes and replaced them with one span event per candidate model instead of a dedicated span per model (to avoid multiplying span count by candidate count on every scored request). Each event is named `score` with `llm_d.scorer.target`/`llm_d.scorer.score` attributes, so opening a single scorer span shows the full per-model breakdown inline in Jaeger/Tempo's event timeline, no extra spans, same drill-down result.

Combined with `llm_d.picker.selected_model` (already set on the picker span) and `llm_d.model_selector.selected_model` (on the top-level `model_selector` span), you get both the winner and the full score breakdown for any picker, not just max-picker.

Let me know if you'd still prefer dedicated child spans over events, happy to switch if that's a stronger convention for `llm-d-router` too.

gyliu513:
@nirrozenbaum sorry for late, but all comments are addressed, can you help check? Thanks

PR #166, opened 2026-06-15
feat: add WithCustomControllers for registering controllers outside plugin chain

yossiovadia:
## Summary

Add a `WithCustomControllers` option to the Runner for registering controller-runtime controllers directly on the manager, outside the plugin lifecycle.

**Problem:** Downstream consumers (e.g. [ai-gateway-payload-processing](https://github.com/opendatahub-io/ai-gateway-payload-processing)) need to register control-plane controllers that create infrastructure resources (Services, HTTPRoutes, Istio resources) from CRDs. Currently, the only way to register controllers is inside a plugin via the Handle's `ReconcilerBuilder`. However, plugins are data-plane components — adding `Owns()` watchers for Services and HTTPRoutes inside a plugin adds heavy informer caches that interfere with the ext-proc gRPC server startup and responsiveness.

**Solution:** `WithCustomControllers` accepts one or more `func(ctrl.Manager) error` setup functions that run after manager creation and before plugin construction. This follows the existing `With*` pattern (`WithExecutableName`, `WithCustomCollectors`).

```go
runner.NewRunner().
    WithExecutableName("my-processor").
    WithCustomControllers(
        func(mgr ctrl.Manager) error {
            return ctrl.NewControllerManagedBy(mgr).
                For(&MyResource{}).
                Owns(&corev1.Service{}).
                Complete(&MyReconciler{Client: mgr.GetClient()})
        },
    ).
    Run(ctx)
```

## Changes

- `Runner` struct: new `customControllerSetupFuncs` field
- `WithCustomControllers()`: functional option method
- `Run()`: executes setup funcs after manager creation, before plugin loading

20 lines added, 1 file changed.

## Test plan

- [x] `make test` — all tests pass
- [x] `make lint` — 0 issues
- [x] `go build ./...` — compiles

## References

- opendatahub-io/ai-gateway-payload-processing#310 — blocked by this framework gap
- Discussion with @nirrozenbaum about separating control-plane vs data-plane controller registration

### DIFF
--- a/cmd/runner/runner.go
+++ b/cmd/runner/runner.go
@@ -29,7 +29,9 @@ import (
 	healthPb "google.golang.org/grpc/health/grpc_health_v1"
 	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
+	ctrlbuilder "sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
@@ -66,6 +68,7 @@ func NewRunner() *Runner {
 		payloadProcessorExecutableName: "payload-processor",
 		profiles:                       map[string]*requesthandling.Profile{},
 		customCollectors:               []prometheus.Collector{},
+		customControllers:              []func(client.Client, *ctrlbuilder.Builder) error{},
 	}
 }
 
@@ -77,8 +80,9 @@ type Runner struct {
 	// profiles is the set of named profiles loaded from the configuration
 	profiles map[string]*requesthandling.Profile
 
-	customCollectors []prometheus.Collector
-	processor        datasource.DatalayerProcessor
+	customCollectors  []prometheus.Collector
+	customControllers []func(client.Client, *ctrlbuilder.Builder) error
+	processor         datasource.DatalayerProcessor
 }
 
 // WithExecutableName sets the name of the executable containing the runner.
@@ -91,6 +95,18 @@ func (r *Runner) WithExecutableName(exeName string) *Runner {
 // WithCustomCollectors sets custom prometheus metrics collectors
 func (r *Runner) WithCustomCollectors(collectors ...prometheus.Collector) *Runner {
 	r.customCollectors = collectors
+	return r
+}
+
+// WithCustomControllers registers controller setup functions that run outside
+// the plugin chain. Use this for control-plane controllers that create
+// infrastructure resources (Services, HTTPRoutes, etc.) and should not be
+// coupled to data-plane plugin lifecycle.
+// Each function receives a client (for scheme access and reconciler construction)
+// and a builder scoped to the manager — matching the Handle.ReconcilerBuilder()
+// pattern used by plugins.
+func (r *Runner) WithCustomControllers(setupFuncs ...func(client.Client, *ctrlbuilder.Builder) error) *Runner {
+	r.customControllers = setupFuncs
 	return r
 }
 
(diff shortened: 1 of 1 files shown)


cmd/runner/runner.go
```diff
@@ -178,6 +189,13 @@ func (r *Runner) Run(ctx context.Context) error {
 		}
 	}
 
+	for _, setupFunc := range r.customControllerSetupFuncs {
+		if err := setupFunc(mgr); err != nil {
+			setupLog.Error(err, "Failed to register custom controller")
```
shaneutt, line 196: It may be worth adding some information to indicate _which_ setup function failed

cmd/runner/runner.go
```diff
@@ -66,6 +66,7 @@ func NewRunner() *Runner {
 		payloadProcessorExecutableName: "payload-processor",
 		profiles:                       map[string]*requesthandling.Profile{},
 		customCollectors:               []prometheus.Collector{},
+		customControllerSetupFuncs:     []func(ctrl.Manager) error{},
```
nirrozenbaum: I would like to avoid giving a consumer of IPP a way to modify controller manager,
while func this signature allows doing custom logic to the controller manager.

can we do here something like the following:
```go
import ctrlbuilder "sigs.k8s.io/controller-runtime/pkg/builder"

customControllers: []func(*ctrlbuilder.Builder) error{},
```

then I'd expect in a consumer to see code like:
```go
func controllerBuilder(builder *ctrlbuilder.Builder) error {
builder().
For(&corev1.ConfigMap{}).
WithEventFilter(predicate()).
Complete(configMapReconciler)
}

...
...
// and in main.go of the custom code:
runner.WithCustomControllers(controllerBuilder).
```

cmd/runner/runner.go
```diff
@@ -178,6 +189,13 @@ func (r *Runner) Run(ctx context.Context) error {
 		}
 	}
 
+	for i, setupFunc := range r.customControllerSetupFuncs {
+		if err := setupFunc(mgr); err != nil {
+			setupLog.Error(err, "Failed to register custom controller", "index", i)
```
nirrozenbaum: I know the index logging was in response to a review comment, but to be honest I don't think this gives any additional information when reading the log. 
we don't see in the log the initial order, so we don't know what does this index mean.

cmd/runner/runner.go
```diff
@@ -178,6 +189,13 @@ func (r *Runner) Run(ctx context.Context) error {
 		}
 	}
 
+	for i, setupFunc := range r.customControllerSetupFuncs {
+		if err := setupFunc(mgr); err != nil {
```
nirrozenbaum:
```go
for _, customControllerFunc := range r.customControllers {
    controllerBuilder := ctrl.NewControllerManagedBy(mgr)
    if err := customControllerFunc(controllerBuider); err != nil {
...
...
}
```

cmd/runner/runner.go
yossiovadia: Dropped the index per Nir's feedback — it wasn't useful without knowing the initial order. Simplified to just the error message.

cmd/runner/runner.go
yossiovadia: Updated to `func(client.Client, *ctrlbuilder.Builder) error`. Runner creates the builder via `ctrl.NewControllerManagedBy(mgr)` — consumer never touches the manager. Client is needed for scheme registration and reconciler construction, matching the Handle pattern.

cmd/runner/runner.go
yossiovadia: Agreed, removed.

cmd/runner/runner.go
yossiovadia: Done — using this pattern with `client.Client` added per your suggestion.

cmd/runner/runner.go
nirrozenbaum: it would be in the responsibility of the custom controller owner to return an error identifying the controller, e.g., sommething like:
```go
return errors.New("failed to register configmap controller")
```

cmd/runner/runner.go
```diff
@@ -94,6 +98,18 @@ func (r *Runner) WithCustomCollectors(collectors ...prometheus.Collector) *Runne
 	return r
 }
 
+// WithCustomControllers registers controller setup functions that run outside
+// the plugin chain. Use this for control-plane controllers that create
+// infrastructure resources (Services, HTTPRoutes, etc.) and should not be
+// coupled to data-plane plugin lifecycle.
+// Each function receives a client (for scheme access and reconciler construction)
+// and a builder scoped to the manager — matching the Handle.ReconcilerBuilder()
+// pattern used by plugins.
```
nirrozenbaum:
```suggestion
// WithCustomControllers registers custom controllers within the 
// controller manager. Use this for control-plane controllers that create
// infrastructure resources (Services, HTTPRoutes, etc.) and should not be
// coupled to data-plane plugin lifecycle.
```

PR #167, opened 2026-06-15
docs: design proposal 044 — plugin body mode capabilities

noyitz:
## Summary

Design proposal for plugins to declare their body mode requirements (NONE / CHUNKED / BUFFERED), enabling the framework to compute the optimal ext_proc `processing_mode` and validate it against the Envoy configuration at startup.

This addresses the root cause of the streaming + metering incompatibility: today `response_body_mode` is a global setting that forces all plugins into the same mode. Plugins that only need the final SSE event (like metering) shouldn't force full buffering that breaks client-side streaming.

## Key ideas

- New `BodyModeCapability` interface — plugins declare what they need
- New `ChunkProcessor` interface — plugins that can handle streaming chunks
- Framework computes effective mode and validates against Envoy config
- **No auto-override** — framework warns/errors, operator decides
- Backward compatible — plugins that don't declare default to BUFFERED

## Context

- [PR #138](https://github.com/llm-d/llm-d-inference-payload-processor/pull/138) — SSE streaming support (prerequisite)
- [RHOAIENG-68300](https://redhat.atlassian.net/browse/RHOAIENG-68300) — streaming broken in 3.4
- The accumulation pattern in `server.go` dates back to the original BBR framework (Mar 2025)

Looking for design feedback before implementation.

### DIFF
--- a/docs/proposals/044-plugin-body-mode-capabilities/README.md
+++ b/docs/proposals/044-plugin-body-mode-capabilities/README.md
@@ -1,0 +1,261 @@
+# Plugin Body Mode Capabilities
+
+Author(s): @noyitz
+
+## Proposal Status
+***Proposed***
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


noyitz:
Closing — per Nir's feedback, the proposal doc is not needed. Implementation is tracked in PRs #169 and #170.

PR #169, opened 2026-06-15
feat: response body mode framework for streaming support

noyitz:
## Summary

- Add `BodyProcessingMode` type (`Skip`, `Chunks`, `Full`) and `ResponseBodyRequirement` interface for response plugins to declare their body access needs
- Pre-compute `NeedsResponseBuffering` per profile at startup based on plugin declarations
- When `NeedsResponseBuffering=false` for the selected profile, ack each response body chunk immediately — enabling real-time client streaming
- Plugins that don't implement the interface default to `Full` (backward compatible)

## Body Processing Modes

| Mode | Behavior | Use case |
|------|----------|----------|
| `Skip` | Plugin doesn't need response body | Headers-only plugins |
| `Chunks` | Plugin processes chunks as they stream (must implement `ChunkProcessor`) | Metering, logging |
| `Full` | Framework buffers entire response before calling plugin | API translation, body mutation |

## How it works

At startup, the config loader iterates each profile's response plugins and checks their `BodyProcessingMode()` declaration:
- If ANY plugin declares `Full` → `NeedsResponseBuffering=true` for that profile
- Otherwise → `NeedsResponseBuffering=false`, chunks are acked immediately

This is a per-profile decision, so different profiles can have different buffering behavior.

## Test plan
- [x] Unit tests for `computeResponseBuffering` covering all mode combinations
- [x] Tests for validation: `Chunks` without `ChunkProcessor` implementation fails at startup
- [x] Multi-profile test verifying independent buffering decisions

### DIFF
--- a/pkg/config/loader/configloader.go
+++ b/pkg/config/loader/configloader.go
@@ -214,9 +214,7 @@ func buildProfiles(rawProfiles []configapi.Profile, handle plugin.Handle) (map[s
 			return nil, fmt.Errorf("the profile %s must have one or both of the Request and Response sections", rawProfile.Name)
 		}
 
-		theProfile := requesthandling.Profile{
-			ResponsePlugins: make([]requesthandling.ResponseProcessor, len(rawProfile.Plugins.Response)),
-		}
+		theProfile := requesthandling.Profile{}
 
 		for _, pluginRef := range rawProfile.Plugins.Request {
 			rawPlugin := handle.Plugin(pluginRef.PluginRef)
@@ -245,17 +243,25 @@ func buildProfiles(rawProfiles []configapi.Profile, handle plugin.Handle) (map[s
 			}
 		}
 
-		for idx, pluginRef := range rawProfile.Plugins.Response {
+		for _, pluginRef := range rawProfile.Plugins.Response {
 			rawPlugin := handle.Plugin(pluginRef.PluginRef)
 			if rawPlugin == nil {
 				return nil, fmt.Errorf("there is no plugin named %s", pluginRef.PluginRef)
 			}
-			thePlugin, ok := rawPlugin.(requesthandling.ResponseProcessor)
-			if !ok {
-				return nil, fmt.Errorf("the plugin named %s is not a ResponseProcessor", pluginRef.PluginRef)
+			matched := false
+			if bodyPlugin, ok := rawPlugin.(requesthandling.ResponseProcessor); ok {
+				theProfile.ResponsePlugins = append(theProfile.ResponsePlugins, bodyPlugin)
+				matched = true
 			}
-			theProfile.ResponsePlugins[idx] = thePlugin
+			if chunkPlugin, ok := rawPlugin.(requesthandling.ResponseChunkProcessor); ok {
+				theProfile.ResponseChunkProcessors = append(theProfile.ResponseChunkProcessors, chunkPlugin)
+				matched = true
+			}
+			if !matched {
+				return nil, fmt.Errorf("the plugin named %s is not a ResponseProcessor or ResponseChunkProcessor", pluginRef.PluginRef)
+			}
 		}
+		theProfile.NeedsResponseBuffering = len(theProfile.ResponsePlugins) > 0
 
 		profiles[rawProfile.Name] = &theProfile
 	}
--- a/pkg/config/loader/configloader.go
+++ b/pkg/config/loader/configloader.go
@@ -248,18 +248,18 @@ func buildProfiles(rawProfiles []configapi.Profile, handle plugin.Handle) (map[s
 			if rawPlugin == nil {
 				return nil, fmt.Errorf("there is no plugin named %s", pluginRef.PluginRef)
 			}
-			matched := false
 			if bodyPlugin, ok := rawPlugin.(requesthandling.ResponseProcessor); ok {
 				theProfile.ResponsePlugins = append(theProfile.ResponsePlugins, bodyPlugin)
-				matched = true
+				continue
 			}
 			if chunkPlugin, ok := rawPlugin.(requesthandling.ResponseChunkProcessor); ok {
 				theProfile.ResponseChunkProcessors = append(theProfile.ResponseChunkProcessors, chunkPlugin)
-				matched = true
+				continue
 			}
-			if !matched {
-				return nil, fmt.Errorf("the plugin named %s is not a ResponseProcessor or ResponseChunkProcessor", pluginRef.PluginRef)
-			}
+			return nil, fmt.Errorf("the plugin named %s is not a ResponseProcessor nor ResponseChunkProcessor", pluginRef.PluginRef)
+		}
+		if len(theProfile.ResponsePlugins) > 0 && len(theProfile.ResponseChunkProcessors) > 0 {
+			return nil, fmt.Errorf("profile %s mixes ResponseProcessor and ResponseChunkProcessor plugins — a profile must use one type exclusively", rawProfile.Name)
 		}
 		theProfile.NeedsResponseBuffering = len(theProfile.ResponsePlugins) > 0
 
--- a/pkg/framework/interface/requesthandling/plugins.go
+++ b/pkg/framework/interface/requesthandling/plugins.go
@@ -52,11 +52,12 @@ type ResponseProcessor interface {
 }
 
 // ResponseChunkProcessor processes individual response body chunks as they
-// stream through without buffering. The framework calls ProcessResponseChunk
-// for each chunk; the returned bytes are forwarded to the client.
+// stream through without buffering. The framework converts the raw chunk bytes
+// to a string once and passes it to all chunk processors. Plugins receive the
+// InferenceResponse to allow header mutation.
 type ResponseChunkProcessor interface {
 	plugin.Plugin
-	ProcessResponseChunk(ctx context.Context, cycleState *plugin.CycleState, chunk []byte, isFinal bool) ([]byte, error)
+	ProcessResponseChunk(ctx context.Context, cycleState *plugin.CycleState, response *InferenceResponse, chunk string, isFinal bool) error
 }
 
 type PostProcessor interface {
--- a/pkg/framework/interface/requesthandling/plugins.go
+++ b/pkg/framework/interface/requesthandling/plugins.go
@@ -43,11 +43,20 @@ type RequestProcessor interface {
 	ProcessRequest(ctx context.Context, cycleState *plugin.CycleState, request *InferenceRequest) error
 }
 
+// ResponseProcessor processes the complete buffered response body.
+// If any plugin in a profile implements this interface, the framework buffers
+// the entire response before calling ProcessResponse on each such plugin.
 type ResponseProcessor interface {
 	plugin.Plugin
-	// ProcessResponse runs the ResponseProcessor plugin.
-	// ResponseProcessor can mutate the headers and/or the body of the response.
 	ProcessResponse(ctx context.Context, cycleState *plugin.CycleState, response *InferenceResponse) error
+}
+
+// ResponseChunkProcessor processes individual response body chunks as they
+// stream through without buffering. The framework calls ProcessResponseChunk
+// for each chunk; the returned bytes are forwarded to the client.
+type ResponseChunkProcessor interface {
+	plugin.Plugin
+	ProcessResponseChunk(ctx context.Context, cycleState *plugin.CycleState, chunk []byte, isFinal bool) ([]byte, error)
 }
 
 type PostProcessor interface {
--- a/pkg/handlers/response.go
+++ b/pkg/handlers/response.go
@@ -136,7 +136,7 @@ func (s *Server) generateEmptyResponseBodyResponse(responseBodyBytes []byte) []*
 func (s *Server) HandleResponseChunk(ctx context.Context, reqCtx *RequestContext, chunkBytes []byte, endOfStream bool) ([]*eppb.ProcessingResponse, error) {
 	// Bodiless requests (e.g., GET /v1/models) may not have a profile set.
 	if reqCtx.Profile == nil || len(reqCtx.Profile.ResponseChunkProcessors) == 0 {
-		return s.buildStreamedChunkResponse(chunkBytes, endOfStream), nil
+		return s.buildStreamedChunkResponse(reqCtx, chunkBytes, endOfStream), nil
 	}
 
 	logger := log.FromContext(ctx).V(logutil.DEFAULT)
@@ -148,7 +148,7 @@ func (s *Server) HandleResponseChunk(ctx context.Context, reqCtx *RequestContext
 		return nil, err
 	}
 
-	return s.buildStreamedChunkResponse(chunkBytes, endOfStream), nil
+	return s.buildStreamedChunkResponse(reqCtx, chunkBytes, endOfStream), nil
 }
 
 // runResponseChunkProcessors executes chunk processors in the order they were registered.
@@ -171,7 +171,9 @@ func (s *Server) runResponseChunkProcessors(ctx context.Context, cycleState *plu
 }
 
 // buildStreamedChunkResponse wraps a chunk in the ext_proc streaming response format.
-func (s *Server) buildStreamedChunkResponse(chunk []byte, endOfStream bool) []*eppb.ProcessingResponse {
+// On the first call (responseHeadersSent=false), it prepends a HeadersResponse to answer
+// the deferred response headers — envoy requires this before it accepts body responses.
+func (s *Server) buildStreamedChunkResponse(reqCtx *RequestContext, chunk []byte, endOfStream bool) []*eppb.ProcessingResponse {
 	responses := []*eppb.ProcessingResponse{
 		{
 			Response: &eppb.ProcessingResponse_ResponseBody{
@@ -191,13 +193,14 @@ func (s *Server) buildStreamedChunkResponse(chunk []byte, endOfStream bool) []*e
 		},
 	}
 
-	if endOfStream {
+	if !reqCtx.ResponseHeadersSent {
 		headerResp := &eppb.ProcessingResponse{
 			Response: &eppb.ProcessingResponse_ResponseHeaders{
 				ResponseHeaders: &eppb.HeadersResponse{},
 			},
 		}
 		responses = append([]*eppb.ProcessingResponse{headerResp}, responses...)
+		reqCtx.ResponseHeadersSent = true
 	}
 
 	return responses
--- a/pkg/handlers/response.go
+++ b/pkg/handlers/response.go
@@ -133,24 +133,45 @@ func (s *Server) generateEmptyResponseBodyResponse(responseBodyBytes []byte) []*
 
 // HandleResponseChunk runs ResponseChunkProcessors on a single response body chunk
 // and wraps the result in the ext_proc streaming response format.
-func (s *Server) HandleResponseChunk(ctx context.Context, reqCtx *RequestContext, chunk []byte, endOfStream bool) ([]*eppb.ProcessingResponse, error) {
+func (s *Server) HandleResponseChunk(ctx context.Context, reqCtx *RequestContext, chunkBytes []byte, endOfStream bool) ([]*eppb.ProcessingResponse, error) {
+	// Bodiless requests (e.g., GET /v1/models) may not have a profile set.
+	if reqCtx.Profile == nil || len(reqCtx.Profile.ResponseChunkProcessors) == 0 {
+		return s.buildStreamedChunkResponse(chunkBytes, endOfStream), nil
+	}
+
+	logger := log.FromContext(ctx).V(logutil.DEFAULT)
+
+	chunk := string(chunkBytes)
+
+	if err := s.runResponseChunkProcessors(ctx, reqCtx.CycleState, reqCtx.Response, chunk, endOfStream, reqCtx.Profile.ResponseChunkProcessors); err != nil {
+		logger.Error(err, "Failed to run response chunk processors")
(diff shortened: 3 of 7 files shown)


docs/proposals/044-plugin-body-mode-capabilities/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+The ext_proc `response_body_mode` is set to `FULL_DUPLEX_STREAMED` in deployment. The IPP framework accumulates all response body chunks in memory before processing ([server.go](../../../pkg/handlers/server.go)), which means:
+
+1. **Streaming is broken** — clients see all response chunks arrive at once after the full response is collected
+2. **Memory is unbounded** — each concurrent request accumulates the full response body with no upper limit
+3. **All plugins pay the cost** — even profiles where no plugin needs the response body force full buffering
+
+## Design
+
+### Response Body Mode Declaration
```
nirrozenbaum: I'd like to propose an alternative. 
instead of making plugins declare if they require full body or can operate on chunks, I propose the following:

go to the file `pkg/framework`interface`requesthandling`. split the ResponseProcessor interface into two interfaces:
```go
type ResponseBodyProcessor interface {
	plugin.Plugin
	// ProcessResponseBody runs the ResponseBodyProcessor plugin.
	// ResponseBodyProcessor can mutate the headers and/or the body of the response.
	ProcessResponseBody(ctx context.Context, cycleState *plugin.CycleState, response *InferenceResponse) error
}

type ResponseChunkProcessor interface {
	plugin.Plugin
	// ProcessResponseChunk runs the ResponseChunkProcessor plugin.
	// ResponseChunkProcessor can operate on a single chunk and can mutate the headers and/or the body of the response.
	ProcessResponseChunk(ctx context.Context, cycleState *plugin.CycleState, responseChunk *InferenceResponseChunk) error // InferenceRequestChunk doesn't exist. need to see if the response can be reused or a new type is needed.
```

then, Profile should be extended to include both `ResponseChunkProcessor[]` AND `ResponseBodyProcessor[]`.

pkg/config/loader/configloader.go
```diff
@@ -96,6 +96,8 @@ func LoadConfiguration(configBytes []byte, handle plugin.Handle, processor datas
 		return nil, err
 	}
 
+	computeResponseBuffering(profiles, logger)
```
nirrozenbaum: this doesn't require a separate function. if you follow the proposal I've added, this can be folded into `buildProfiles` function, where we need to maintain for each profile a boolean field `RequireResponseBuffering: bool` (which is actually a result of `if len(ResponseBodyProcessor) > 0`, so maybe we don't even need the extra field).

docs/proposals/044-plugin-body-mode-capabilities/README.md
```diff
@@ -0,0 +1,90 @@
...
+```
+INFO  Profile response buffering computed  profile=default needsResponseBuffering=true bufferingPlugins=[api-translation]
+INFO  Profile response buffering computed  profile=headers-only needsResponseBuffering=false bufferingPlugins=[]
+```
+
+Plugins that don't implement `ResponseBodyRequirement` get a warning:
+```
+INFO  Response plugin does not declare ResponseBodyRequirement, defaulting to BodyFull  profile=default plugin=legacy-plugin/legacy
+```
+
```
nirrozenbaum: we don't need this. llm-d IPP is a new repo and hasn't been through the process of a formal release yet. so we don't need to support backward compatibility, which can simplify a bit.

docs/proposals/044-plugin-body-mode-capabilities/README.md
```diff
@@ -0,0 +1,90 @@
...
+| 2 — ChunkProcessor | `ChunkProcessor` interface for `BodyChunked` plugins, per-chunk dispatch in server loop | Next PR |
+| 3 — Plugin declarations | Existing plugins implement `ResponseBodyRequirement` (ODH repo) | Separate PR |
+
+## Migration
+
+Existing plugins that don't implement `ResponseBodyRequirement` default to `BodyFull`. No breaking changes — everything works exactly as before. Profiles with only legacy plugins will continue to buffer.
+
+## Future: Dynamic Mode Override
+
+Envoy ext_proc supports `mode_override` in responses. A future enhancement could override `response_body_mode` to `NONE` per-request when no response plugin in the selected profile needs the body at all. This is out of scope for the initial implementation.
```
nirrozenbaum: this is very interesting. I didn't know this option exist. 
can you open an issue about this so we can track it and make sure it goes into next version? 
100% agree this is not first priority.

docs/proposals/044-plugin-body-mode-capabilities/README.md
noyitz: Done — reworked the PR to follow your approach. Split `ResponseProcessor` into:
- `ResponseProcessor` — processes full buffered body (existing interface, unchanged)
- `ResponseChunkProcessor` — processes individual chunks without buffering (new)

The config loader sorts response plugins into the right list based on which interface they implement. Profile now has `ResponsePlugins`, `ResponseChunkProcessors`, and `NeedsResponseBuffering` (computed automatically from `len(ResponsePlugins) > 0`).

Removed the `BodyProcessingMode` enum and `ResponseBodyRequirement` interface entirely. Proposal doc removed too.

pkg/config/loader/configloader.go
noyitz: Done — folded into `buildProfiles`. The buffering flag is computed inline: `theProfile.NeedsResponseBuffering = len(theProfile.ResponsePlugins) > 0`

docs/proposals/044-plugin-body-mode-capabilities/README.md
noyitz: Removed. No backward compat shims.

docs/proposals/044-plugin-body-mode-capabilities/README.md
noyitz: Will open a tracking issue for the FULL_DUPLEX_STREAMED per-chunk envoy mode.

pkg/handlers/server.go
```diff
@@ -176,15 +176,39 @@ func (s *Server) Process(srv extProcPb.ExternalProcessor_ProcessServer) error {
...
-			responseBody = append(responseBody, v.ResponseBody.Body...)
-			if !v.ResponseBody.EndOfStream {
-				continue
+
+			hasChunkProcessors := reqCtx.Profile != nil && len(reqCtx.Profile.ResponseChunkProcessors) > 0
+			needsBuffering := !hasChunkProcessors
+			if reqCtx.Profile != nil && reqCtx.Profile.NeedsResponseBuffering {
+				needsBuffering = true
+			}
+			if needsBuffering {
```
nirrozenbaum: this is wrong. in case the profile doesn't contain any response plugin at all we want to stream the chunks and not buffer them.
the check should be the other way around - 
if the profile has any ResponseProcessor plugin, it requires buffering. and that was already calculated in configloader (L264). 
so this logic can be replaced with (we can assume profile was set as part of the request handling):

```suggestion
			if reqCtx.Profile.NeedsResponseBuffering {
```

pkg/config/loader/configloader.go
```diff
@@ -245,17 +243,25 @@ func buildProfiles(rawProfiles []configapi.Profile, handle plugin.Handle) (map[s
...
+				theProfile.ResponsePlugins = append(theProfile.ResponsePlugins, bodyPlugin)
+				matched = true
+			}
+			if chunkPlugin, ok := rawPlugin.(requesthandling.ResponseChunkProcessor); ok {
+				theProfile.ResponseChunkProcessors = append(theProfile.ResponseChunkProcessors, chunkPlugin)
+				matched = true
+			}
+			if !matched {
+				return nil, fmt.Errorf("the plugin named %s is not a ResponseProcessor or ResponseChunkProcessor", pluginRef.PluginRef)
 			}
```
nirrozenbaum: we don't need `matched` variable (extra vars are adding overhead to a reader). 

```suggestion
			if bodyPlugin, ok := rawPlugin.(requesthandling.ResponseProcessor); ok {
				theProfile.ResponsePlugins = append(theProfile.ResponsePlugins, bodyPlugin)
				continue
			}
			if chunkPlugin, ok := rawPlugin.(requesthandling.ResponseChunkProcessor); ok {
				theProfile.ResponseChunkProcessors = append(theProfile.ResponseChunkProcessors, chunkPlugin)
				continue
			}
			return nil, fmt.Errorf("the plugin named %s is not a ResponseProcessor nor ResponseChunkProcessor", pluginRef.PluginRef)
```

pkg/handlers/server.go
```diff
@@ -176,15 +176,39 @@ func (s *Server) Process(srv extProcPb.ExternalProcessor_ProcessServer) error {
...
+				chunk := v.ResponseBody.Body
+				isFinal := v.ResponseBody.EndOfStream
+				if reqCtx.Profile != nil {
+					for _, cp := range reqCtx.Profile.ResponseChunkProcessors {
+						chunk, err = cp.ProcessResponseChunk(ctx, reqCtx.CycleState, chunk, isFinal)
+						if err != nil {
+							break
+						}
+					}
+				}
```
nirrozenbaum: this code doesn't belong here.. should be moved to response.go into a dedicated HandleResponseChunk function.
also we can remove the isFinal and chunk variables. the function can get `v.ResponseBody.Body` (call the var in the function input `chunk` and  `v.ResponseBody.EndOfStream`)

pkg/handlers/response.go
```diff
@@ -131,6 +131,56 @@ func (s *Server) generateEmptyResponseBodyResponse(responseBodyBytes []byte) []*
 	return responses
 }
 
+// buildChunkPassthroughResponse wraps a single response body chunk in the
+// ext_proc streaming response format. Used when no ResponseProcessor (buffered)
+// plugins are present — chunks flow through ResponseChunkProcessors and are
+// forwarded immediately.
+func (s *Server) buildChunkPassthroughResponse(chunk []byte, isFinal bool) []*eppb.ProcessingResponse {
```
nirrozenbaum: why do you do passthrough? what about calling the RespnoseChunkProcessor plugins?
you need to create `HandleResponseChunk` function in this file, that follows the same principles of `HandleResponseBody` function. 

also, need to make sure we don't return response from both FullBody and last chunk handler functions. otherwise the result may be not what we intended.

a simple approach to start with could be to limit a profile to include either no response plugins, or only chunk processors or only response processors. (in other words, don't allow mix of full/chunk processors within the scope of a single profile)

pkg/handlers/server.go
noyitz: Done — simplified to `reqCtx.Profile.NeedsResponseBuffering`. No-plugin profiles now stream without buffering.

pkg/config/loader/configloader.go
noyitz: Done — removed `matched`, using `continue` pattern.

pkg/handlers/server.go
noyitz: Done — moved to `HandleResponseChunk` in response.go. Runs chunk processors with metrics and logging, same pattern as `HandleResponseBody`.

pkg/handlers/response.go
noyitz: Done — added validation in config loader: profiles cannot mix ResponseProcessor and ResponseChunkProcessor. A profile uses one type exclusively. Per our Slack discussion, the profile picker handles passthrough-vs-translation at runtime using separate profiles.

pkg/handlers/server_test.go
```diff
@@ -123,6 +123,7 @@ func TestHandleResponseBody_Streaming(t *testing.T) {
 	wantFullBody := []byte(`{"choices":[{"text":"Hello!"}]}`)
 
 	profiles := newTestProfiles()
+	profiles[testProfileName].NeedsResponseBuffering = true
```
nirrozenbaum, line 126: can we add in a follow up unit tests to cover the new funcionality?
- mix of body response plugins + chunk response fails.
- setting profile with chunks only perform streaming.
- setting profile with full only is buffering.
- setting profile with no plugins performs streaming.

not a blocker for this PR (can be pushed in follow up).

pkg/handlers/server.go
```diff
@@ -176,15 +176,24 @@ func (s *Server) Process(srv extProcPb.ExternalProcessor_ProcessServer) error {
...
+				responseBody = append(responseBody, v.ResponseBody.Body...)
+				if !v.ResponseBody.EndOfStream {
+					continue
+				}
+				reqCtx.ResponseCompleteTimestamp = time.Now()
+				model, _ := reqCtx.Request.Body["model"].(string)
+				metrics.RecordRequestTTFT(model, reqCtx.ResponseFirstChunkTimestamp.Sub(reqCtx.RequestReceivedTimestamp))
+				responses, err = s.HandleResponseBody(ctx, reqCtx, responseBody)
+				loggerVerbose.Info("processing response body complete")
+			} else {
```
nirrozenbaum, line 189: now this looks nice and clean :)

pkg/handlers/server.go
```diff
@@ -176,15 +176,24 @@ func (s *Server) Process(srv extProcPb.ExternalProcessor_ProcessServer) error {
...
-				continue
+
+			if reqCtx.Profile.NeedsResponseBuffering {
+				responseBody = append(responseBody, v.ResponseBody.Body...)
+				if !v.ResponseBody.EndOfStream {
+					continue
+				}
+				reqCtx.ResponseCompleteTimestamp = time.Now()
+				model, _ := reqCtx.Request.Body["model"].(string)
+				metrics.RecordRequestTTFT(model, reqCtx.ResponseFirstChunkTimestamp.Sub(reqCtx.RequestReceivedTimestamp))
```
nirrozenbaum: can you move these two lines after the else block (but still inside the case) so that we report TTFT metric in both cases if we handle chunks or handle full response?

nirrozenbaum: this would require only to update the `continue` in L183.

pkg/framework/interface/requesthandling/plugins.go
```diff
@@ -43,13 +43,22 @@ type RequestProcessor interface {
...
-	// ResponseProcessor can mutate the headers and/or the body of the response.
 	ProcessResponse(ctx context.Context, cycleState *plugin.CycleState, response *InferenceResponse) error
 }
 
+// ResponseChunkProcessor processes individual response body chunks as they
+// stream through without buffering. The framework calls ProcessResponseChunk
+// for each chunk; the returned bytes are forwarded to the client.
+type ResponseChunkProcessor interface {
+	plugin.Plugin
+	ProcessResponseChunk(ctx context.Context, cycleState *plugin.CycleState, chunk []byte, isFinal bool) ([]byte, error)
```
nirrozenbaum: the framework should convert the chunkBytes once to string and call the various chunk plugins with string. otherwise every plugin will have to pay to conversion overhead. 
also, this interface should get the InferenceResponse, otherwise it cannot mutate headers.

pkg/handlers/response.go
```diff
@@ -131,6 +131,57 @@ func (s *Server) generateEmptyResponseBodyResponse(responseBodyBytes []byte) []*
 	return responses
 }
 
+// HandleResponseChunk runs ResponseChunkProcessors on a single response body chunk
+// and wraps the result in the ext_proc streaming response format.
+func (s *Server) HandleResponseChunk(ctx context.Context, reqCtx *RequestContext, chunk []byte, endOfStream bool) ([]*eppb.ProcessingResponse, error) {
+	logger := log.FromContext(ctx).V(logutil.DEFAULT)
+	verboseLogger := logger.V(logutil.VERBOSE)
+
```
nirrozenbaum, line 166: we should mirror from HandleResponseBody the case when no plugins are configured at all, and also cover the case of bodiless request (like Get v1/models) where profile is not set at all.

```suggestion
// if a bodiless request is sent, profile could be left nil. check to avoid panic with nil exception
if reqCtx.Profile == nil || len(reqCtx.Profile.ResponseChunkProcessors) == 0 {
	// generate empty streamed response and stream back the chunk
}

var chunk string
if err := json.Unmarshal(chunkBytes, &chunk); err != nil {
	logger.Error(err, "Failed to parse response chunk, skipping response chunk plugins")
	// generate empty streamed response and stream back the chunk
}

if err := s.runResponseChunkProcessors(ctx, reqCtx.CycleState, chunk, reqCtx.Profile.ResponseChunkProcessors); err != nil {
	return nil, err
}
...

```

pkg/handlers/server_test.go
noyitz: Agreed — will add in a follow-up PR.

pkg/handlers/server.go
noyitz: Done — moved TTFT metric and ResponseCompleteTimestamp after the if/else block. Both paths now record metrics on EoS.

pkg/framework/interface/requesthandling/plugins.go
noyitz: Done — changed interface to `ProcessResponseChunk(ctx, cycleState, response *InferenceResponse, chunk string, isFinal bool) error`. Framework converts bytes to string once. Response passed for header mutation.

pkg/handlers/response.go
noyitz: Done — added nil profile check, empty processor check, and extracted `runResponseChunkProcessors` + `buildStreamedChunkResponse` mirroring the body processing pattern.

pkg/handlers/response.go
```diff
@@ -131,6 +131,81 @@ func (s *Server) generateEmptyResponseBodyResponse(responseBodyBytes []byte) []*
 	return responses
 }
 
+// HandleResponseChunk runs ResponseChunkProcessors on a single response body chunk
+// and wraps the result in the ext_proc streaming response format.
+func (s *Server) HandleResponseChunk(ctx context.Context, reqCtx *RequestContext, chunkBytes []byte, endOfStream bool) ([]*eppb.ProcessingResponse, error) {
```
nirrozenbaum, line 136: something is still wrong in this function.. 
chunk processor plugins cannot mutate the chunk content, and even if they do it's ignored in L151. 
also, I've noticed that BodyMutation is always used even if chunk hasn't changed, which is an expensive operation in envoy.. 

we should also probably distinguish between the case when chunk was mutated and the case when it was passed through. 
we can improve that last point on a follow up PR, leaving mutation always works for now. 
but the first point is functionality wise wrong.

noyitz:
Addressed the latest comments in two commits:

**63991fd — chunk mutation via InferenceResponse:**
- Added `CurrentChunk`, `SetChunk()`, `ChunkMutated()`, `ResetChunkState()` on `InferenceResponse` — plugins read/mutate the chunk through the response object, not the pass-by-value string parameter.
- `HandleResponseChunk` calls `ResetChunkState(chunk)` before plugins run, then checks `ChunkMutated()` and uses the modified version in `buildStreamedChunkResponse`.
- `runResponseChunkProcessors` passes `response.CurrentChunk` to each plugin so mutations chain across the pipeline.

**63991fd — TTFT metric fix:**
- Changed `continue` to `break` in the buffering accumulation branch so the TTFT metric code after the if/else block is reachable for both buffered and streaming paths.

**d793ccb — tests:**
- `TestChunkMutation` — verifies `SetChunk` marks mutated, `ResetChunkState` sets chunk without marking mutated.
- `TestChunkMutation_ResetClearsMutatedFlag` — verifies `ResetChunkState` clears the mutated flag between chunks.

Also fixed `buildStreamedChunkResponse` to send `HeadersResponse` with the first chunk (not EoS) — envoy needs it before accepting body responses in FULL_DUPLEX_STREAMED mode.

pkg/handlers/response.go
```diff
@@ -131,6 +131,89 @@ func (s *Server) generateEmptyResponseBodyResponse(responseBodyBytes []byte) []*
...
+func (s *Server) runResponseChunkProcessors(ctx context.Context, cycleState *plugin.CycleState, response *requesthandling.InferenceResponse, isFinal bool, processors []requesthandling.ResponseChunkProcessor) error {
+	logger := log.FromContext(ctx).V(logutil.DEFAULT)
+	verboseLogger := logger.V(logutil.VERBOSE)
+
+	for _, cp := range processors {
+		if verboseLogger.Enabled() {
+			verboseLogger.Info("Executing response chunk plugin", "plugin", cp.TypedName())
+		}
+		before := time.Now()
+		err := cp.ProcessResponseChunk(ctx, cycleState, response, response.CurrentChunk, isFinal)
```
nirrozenbaum, line 172: we should change the inferface and call to remove here `response.Chunk`. this creates ambiguity for plugin implementor who may incorrectly update the chunk content by changing the string. 
passing in response only make the implementer more aware of the fact that working with chunk should be through the response object. read it through CurrentChunk and set it via SetChunk.

fine for a follow up, not a blocker.

nirrozenbaum:
left few non blocking comments for follow up. 

/lgtm
/approve

noyitz:
Addressed in 5b16f0c — removed `chunk string` parameter from `ProcessResponseChunk` interface. Plugins now read/write exclusively through `response.CurrentChunk` and `response.SetChunk()`. No ambiguity.

All tests pass, CI should go green.

nirrozenbaum:
> Addressed in [5b16f0c](https://github.com/llm-d/llm-d-inference-payload-processor/commit/5b16f0c115e0127311a04f1689b42df6d5d6d2c9) — removed `chunk string` parameter from `ProcessResponseChunk` interface. Plugins now read/write exclusively through `response.CurrentChunk` and `response.SetChunk()`. No ambiguity.
> 
> All tests pass, CI should go green.

@noyitz can you push in a new pr? this one already merged

PR #172, opened 2026-06-17
feat: model name Filter

### LINKED ISSUE #171: [Feature]: Add a basic model name Filter

A basic Filter mechanism that gets the restricts the candidate models to the model name(s) requested in the body.

**Why is this needed**:

aradhalevy:
**What type of PR is this?**

Adds a `model-name-filter` that restricts the candidate models to the model name requested in the body.

The configured request-body fieldis read as a string:
- a single model name → that model is the only candidate;
- absent / empty string / `"[]"` → no constraint, all candidates pass through.

Fixes #171 

```release-note
NONE
```


pkg/modelselector/model_selector_pipeline.go
```diff
@@ -146,7 +146,9 @@ func (p *ModelSelectorPipeline) String() string {
 func (p *ModelSelectorPipeline) Run(ctx context.Context, request *requesthandling.InferenceRequest, cycleState *plugin.CycleState, candidateModels []datalayer.Model) (*modelselector.PipelineRunResult, error) {
 	models := p.runFilterPlugins(ctx, request, cycleState, candidateModels)
 	if len(models) == 0 {
-		return nil, errors.New("no models available after filtering")
+		// Typed so the handler maps it to an HTTP 400 ImmediateResponse instead
+		// of failing the ext_proc stream.
+		return nil, errcommon.Error{Code: errcommon.BadRequest, Msg: "no models available after filtering"}
```
nirrozenbaum: why is it a bad request? (I don't think it is).
ResourceExhausted is a better fit here considering the fact that multiple filters can be used and most of them will be around models that the user is not allowed or doesn't have quota.

nirrozenbaum (commented):
not sure I understood the intention in this PR. 
when a model is specified in the body, we shouldn't run model selector at all. 
(when it is a single model). 
this should be done by creating a profile picker that checks whether model is set in the body or not, and if model is already set, then select a profile without model selector. 

regarding the use case of multiple models in an array alike value.. this is an invented syntax, no cli or client library for openai (or other formats) would work that way.

I would suggest to do some changes here:
- if your intention is to restrict to a single model - that should be done using new ProfilePicker. 
- if your intention is to restrict to a subset of models, I'd be happy to get more details on this. assuming we have models a,b,c,d,e and this filter allows only b,c. then why do we need models a,d,e? will those ever be used? why are they deployed? 
- if it's still a valid case - I think it's legit to specify a subset of allowed models in a header, not inside the body. (same is done in epp btw, using subset header).

pkg/modelselector/model_selector_pipeline.go
aradhalevy: Done

aradhalevy:
@nirrozenbaum Changed the Filter to allow Filter the model name exactly, else an empty field and then we let all the possible models work

nirrozenbaum:
> @nirrozenbaum Changed the Filter to allow Filter the model name exactly, else an empty field and then we let all the possible models work

@aradhalevy thanks. this looks better. 
what are your thoughts about generalizing the filter to be called something like `ByField`, with configuration of `fieldName` and `optionalValues`? 
This kind of filter would allow you to configure `ByModel` filter (from type `ByField`, with name `ByModel`) and would also allow a broader behavior.
e.g., (just as an example) - when a request is sent to an external anthropic model on bedrock, we would be able to filter models by their version (cause anthropic request specify a field `"anthropic_version": "bedrock-2023-05-31"`). 

wdyt?

aradhalevy:
@nirrozenbaum This sounds like a good idea, but I think keeping this PR simple to have a working Filter would be better. I suggest opening an issue to extend this Filter later. how about that?

pkg/common/error/error.go
```diff
@@ -47,10 +48,11 @@ func (e Error) Error() string {
...
 }
 
-// CanonicalCode returns the error's ErrorCode.
+// CanonicalCode returns the error's ErrorCode. It unwraps wrapped errors so a
+// typed Error survives an fmt.Errorf("...: %w", err) on the way up.
 func CanonicalCode(err error) string {
-	e, ok := err.(Error)
-	if ok {
+	var e Error
+	if errors.As(err, &e) {
```
nirrozenbaum: is this change related to the PR?

pkg/framework/plugins/modelselector/filter/modelname/filter.go
```diff
@@ -0,0 +1,101 @@
...
+// "model" field. The field must be a string; an absent field or empty string
+// does not constrain the candidates; any non-string value is malformed and
+// yields no candidates (the pipeline rejects the request).
+func (f *ModelNameFilter) Filter(ctx context.Context, _ *plugin.CycleState, request *requesthandling.InferenceRequest, models []datalayer.Model) []datalayer.Model {
+	logger := log.FromContext(ctx)
+
+	requested, ok := request.Body[requestModelField].(string)
+	if !ok && request.Body[requestModelField] != nil {
+		logger.V(logutil.VERBOSE).Info("malformed model field in request body, no available model candidates", "field", requestModelField)
+		return nil
```
nirrozenbaum: let's avoid nil here:

```suggestion
		return []datalayer.Model{}
```

pkg/framework/plugins/modelselector/filter/modelname/README.md
```diff
@@ -0,0 +1,20 @@
...
+model field is treated as a single model name.
+
+It is registered as type `model-name-filter` and runs as a modelselector filter.
+
+The filter matches the requested name against the candidate models the pipeline hands it from the datalayer. "Available" below means "present in that candidate list".
+
+## What it does
+
+1. Reads the `model` field from the request body.
+2. When the field holds an available model name, that one model becomes the candidate.
```
nirrozenbaum:
```suggestion
2. When the field holds an available model name, that one model becomes the single candidate, which de facto means that model is selected (picker doesn't have alternatives).
```

nirrozenbaum:
> @nirrozenbaum This sounds like a good idea, but I think keeping this PR simple to have a working Filter would be better. I suggest opening an issue to extend this Filter later. how about that?

@aradhalevy we can do that. I left few nits if you can please handle and then we should be good to merge.

pkg/common/error/error.go
aradhalevy: Yes. The pipeline now returns a typed `errcommon.Error` (the `ResourceExhausted` on line 151), and modelselector/plugin.go:88 wraps it with `fmt.Errorf("model selection failed: %w", err)` before it reaches the handler's `BuildErrResponse` (server.go:202). The old `err.(Error)` type assertion fails on a wrapped error and returns `Unknown`, which makes the handler fail the ext_proc stream instead of emitting the 429 `ImmediateResponse`.
`errors.As` unwraps it so the typed code survives. Without this change the new filter's 429 path doesn't actually work end-to-end.

pkg/framework/plugins/modelselector/filter/modelname/filter.go
aradhalevy: Done, and also in line 100 in the same way

pkg/framework/plugins/modelselector/filter/modelname/README.md
aradhalevy: Added

pkg/common/error/error.go
nirrozenbaum: not sure I understand. 
modelselector has no file plugin.go, and we don't wrap the error using fmt.Errof before returning an error.
can you please provide a pointer to help me understand?

aradhalevy: Sorry I wasn't clear, the full stack of calls: 
I changed In this PR the `Run` function to return the `errcommon.Error{Code: ResourceExhausted}` error [here](https://github.com/llm-d/llm-d-inference-payload-processor/pull/172/changes#diff-75ed5de8467b8d5c280e8cffe6e63e83c1dc421068b6a61059a3cef4bd400d56R151). 
 
What calls this `Run` function is the `Select` function in `pkg/modelselector/modelselector.go`. It calls it [here](https://github.com/llm-d/llm-d-inference-payload-processor/blob/ac26114049ffcfe17634981515c90dd9f1f5e329/pkg/modelselector/modelselector.go#L67) and passes it through unwrapped.

What calls this `Select` function is the `ProcessRequest` function in `pkg/framework/plugins/requesthandling/modelselector/plugin.go` in line 86, and in [line 88](https://github.com/llm-d/llm-d-inference-payload-processor/blob/main/pkg/framework/plugins/requesthandling/modelselector/plugin.go#L88) where we wrap the error with `%w` using `fmt.Errof`.

This is where the wrapping happens.

As for where we get to `BuildErrResponse`:

In `pkg/handlers/server.go` we call [`HandleRequestBody`](https://github.com/llm-d/llm-d-inference-payload-processor/blob/ac26114049ffcfe17634981515c90dd9f1f5e329/pkg/handlers/server.go#L167). 

`HandleRequestBody` is defined in `pkg/handlers/request.go` and calls `runRequestPlugins` which calls [`ProcessRequest`](https://github.com/llm-d/llm-d-inference-payload-processor/blob/ac26114049ffcfe17634981515c90dd9f1f5e329/pkg/handlers/request.go#L139). 

When an error happens in the `Run` function I changed in this PR it comes all the way to [`HandleRequestBody`](https://github.com/llm-d/llm-d-inference-payload-processor/blob/ac26114049ffcfe17634981515c90dd9f1f5e329/pkg/handlers/server.go#L167) after it was wrapped along the way, and it continues to [`BuildErrResponse`](https://github.com/llm-d/llm-d-inference-payload-processor/blob/ac26114049ffcfe17634981515c90dd9f1f5e329/pkg/handlers/server.go#L202). 

I know this is a long call stack, hope this is clearer now.

shmuelk:
I don't understand this filter. Doesn't it negate the the entire idea of the model selector?

If one wants to force a model either don't include the model selector in the configuration, or "fix" the model selector so that it doesn't always select a model.

pkg/common/error/error.go
nirrozenbaum: haha no worries. I was able to follow.
> What calls this Select function is the ProcessRequest function in pkg/framework/plugins/requesthandling/modelselector/plugin.go in line 86, and in [line 88](https://github.com/llm-d/llm-d-inference-payload-processor/blob/main/pkg/framework/plugins/requesthandling/modelselector/plugin.go#L88) where we wrap the error with %w using fmt.Errof.

can we return the error here unwrapped instead of changing the errors file? 
ideally plugins should not wrap the error with a non built-in error, otherwise they hide the real issue. 
this seems like it was a mistake in `modelselector/plugin.go`.

aradhalevy: Done

nirrozenbaum:
/lgtm
/approve

Thanks 🙏🏼

PR #175, opened 2026-06-17
Wire model-config-datasource into the runner.

Mohammad-nassar10:
**What type of PR is this?**
/kind feature

Register `model-config-datasource` plugin in the runner.

Fixes #

```release-note
NONE
```

### DIFF
--- a/config/charts/payload-processor/templates/config.yaml
+++ b/config/charts/payload-processor/templates/config.yaml
@@ -26,4 +26,11 @@ data:
     kind: PayloadProcessorConfig
     {{- .Values.payloadProcessor.customConfig | toYaml | nindent 4 }}
   {{- end }}
+  {{- if .Values.payloadProcessor.listModels }}
+  {{- $models := list }}
+  {{- range .Values.payloadProcessor.listModels }}
+  {{- $models = append $models (dict "name" .) }}
+  {{- end }}
+  models.json: {{ dict "models" $models | toJson | quote }}
+  {{- end }}
 ---
--- a/config/charts/payload-processor/templates/config.yaml
+++ b/config/charts/payload-processor/templates/config.yaml
@@ -26,11 +26,4 @@ data:
     kind: PayloadProcessorConfig
     {{- .Values.payloadProcessor.customConfig | toYaml | nindent 4 }}
   {{- end }}
-  {{- if .Values.payloadProcessor.listModels }}
-  {{- $models := list }}
-  {{- range .Values.payloadProcessor.listModels }}
-  {{- $models = append $models (dict "name" .) }}
-  {{- end }}
-  models.json: {{ dict "models" $models | toJson | quote }}
-  {{- end }}
 ---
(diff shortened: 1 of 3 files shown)


config/charts/payload-processor/templates/config.yaml
```diff
@@ -26,4 +26,11 @@ data:
     kind: PayloadProcessorConfig
     {{- .Values.payloadProcessor.customConfig | toYaml | nindent 4 }}
   {{- end }}
+  {{- if .Values.payloadProcessor.listModels }}
+  {{- $models := list }}
+  {{- range .Values.payloadProcessor.listModels }}
+  {{- $models = append $models (dict "name" .) }}
+  {{- end }}
+  models.json: {{ dict "models" $models | toJson | quote }}
+  {{- end }}
```
nirrozenbaum: it's not clear what is listModels from this code. 
can you please add the relevant documentation to the readme.MD inside the helm chart dir? 
the optional flags should be documentation so whoever uses the helm can configure IPP easily.

Mohammad-nassar10: Good point, added, thanks!

nirrozenbaum (commented):
@Mohammad-nassar10 thanks for adding the doc. 
now that I understand what was the intention in the PR - this looks somehow very limiting - the list of models is static, there is no way to add a model dynamically (nor remove). it seems like an easy way for a PoC but not the proper way moving forward. 

the payload processing config is intended to be used for plugins configuration. 
seems like this PR is making the config define the source of truth of available models (this is not a plugin, although one could argue data collector is). 

If we take it to EPP's world to compare just to make the point - 
think how it would look like if one would have to configure the available endpoints of EPP through the plugins config file. this sounds incorrect to me.

alternative direction that could work:
add a configmap that stores models. configmap could also add metadata for every model entry (e.g., like its cost per 1M tokens). add a controller that watches the "modelsList" configmap(s) and fills that data into the datastore.

shmuelk:
@nirrozenbaum wrote:

> the list of models is static, there is no way to add a model dynamically (nor remove). it seems like an easy way for a PoC but not the proper way moving forward.

The updates to the README.md state:
> `payloadProcessor.listModels` is a list of model names that the IPP should be aware of. When set, the chart renders a `models.json` file into the ConfigMap, which is mounted at `/config/models.json` inside the pod. The `model-config-datasource` plugin reads this file to populate the IPP datastore and watches it for live updates on ConfigMap remounts.

That is not part of the IPP's configuration ConfigMap. It is a separate ConfigMap. If and how that ConfigMap is checked for changes is a different thing, which should be handled separately from this PR.

ronenkat:
> @Mohammad-nassar10 thanks for adding the doc. now that I understand what was the intention in the PR - this looks somehow very limiting - the list of models is static, there is no way to add a model dynamically (nor remove). it seems like an easy way for a PoC but not the proper way moving forward.
> 
> the payload processing config is intended to be used for plugins configuration. seems like this PR is making the config define the source of truth of available models (this is not a plugin, although one could argue data collector is).
> 
> If we take it to EPP's world to compare just to make the point - think how it would look like if one would have to configure the available endpoints of EPP through the plugins config file. this sounds incorrect to me.
> 
> alternative direction that could work: add a configmap that stores models. configmap could also add metadata for every model entry (e.g., like its cost per 1M tokens). add a controller that watches the "modelsList" configmap(s) and fills that data into the datastore.

We already have a `modelconfigcollector` data source that reads a models-config.json file when ever it changes.
This PR enables IPP to run with a pre-defined list of models.
There could be multiple ways to configure IPP with models, this is just the first simple option.

Mohammad-nassar10:
@nirrozenbaum Keeping only the missing plugin registration in this PR. Will handle the config in separate PR.

PR #176, opened 2026-06-17
fix: strip stream_options from request body before parsing

noyitz:
## Summary
- Strip `stream_options` from request body raw bytes before `json.Unmarshal` in `HandleRequestBody`
- OpenAI SDKs add `stream_options` to streaming requests, but Anthropic rejects it with `400 Extra inputs are not permitted`
- In `FULL_DUPLEX_STREAMED` mode, body mutations via `StreamedBodyResponse` cannot reliably remove fields from chunks already forwarded by envoy — the strip must run on accumulated raw bytes
- Uses JSON parse/delete/re-serialize to avoid false matches when the text "stream_options" appears inside message string values

## Test plan
- [x] `TestStripStreamOptions` — 7 unit test cases: top-level key removal, no-op when absent, text inside strings preserved, large body (50KB), invalid JSON passthrough, empty body
- [x] `TestHandleRequestBody_StripsStreamOptions` — 3 integration tests: small body, 100KB body, body without stream_options
- [x] All existing handler tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/handlers/request.go
+++ b/pkg/handlers/request.go
@@ -64,6 +64,8 @@ func (s *Server) HandleRequestHeaders(ctx context.Context, reqCtx *RequestContex
 func (s *Server) HandleRequestBody(ctx context.Context, reqCtx *RequestContext, requestBodyBytes []byte) ([]*eppb.ProcessingResponse, error) {
 	var ret []*eppb.ProcessingResponse
 
+	requestBodyBytes = stripStreamOptions(requestBodyBytes)
+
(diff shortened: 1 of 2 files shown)


noyitz:
Closing — the framework should not strip fields provider-agnostically. The real fix is the metering type mismatch (ReadCycleStateKey[string] vs APIFormat) in opendatahub-io/ai-gateway-payload-processing#332. Provider-specific field stripping belongs in the api-translation plugin.

PR #178, opened 2026-06-18
Session affinity routing

szedan-rh:
The problem: When a user has a multi-turn conversation, each request might land on a different server. That server has to re-process all the previous context from scratch — wasting GPU time and adding latency.

  The fix: We stick each conversation to one server.

  ---
  How:
  
  1. First request comes in — we give it a session ID (or the client sends one)
  2. We look at which servers are healthy, then use the session ID to pick one
  3. Every future request with that same session ID goes to the same server
  4. The server's cache stays warm — no re-processing

  If a server goes down:

  - Only the conversations that were on that server get moved
  - Everyone else is unaffected
  - The moved conversations lose their cache, but it rebuilds naturally

  That's it. Same session, same server, faster responses.


szedan-rh:
@nirrozenbaum - Could you please review?

szedan-rh:
The PR is large because the html visualization how the flow work.

pkg/framework/plugins/modelselector/filter/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (f *SessionAffinityFilter) TypedName() plugin.TypedName {
+	return f.typedName
+}
+
+// Filter implements modelselector.Filter. It reads or generates a session ID,
+// builds a consistent hash ring from models with weight > 0, and returns the
+// single model that the session hashes to.
+func (f *SessionAffinityFilter) Filter(ctx context.Context, cycleState *plugin.CycleState, request *requesthandling.InferenceRequest, models []datalayer.Model) []datalayer.Model {
```
ronenkat: Why is the session affinity implemented as a filter? and not a scorer?
In the payload processor as opposed to the llm-d router, we choose models (or inference pools), not servers.
Consider the following cases: A user decides to use a different model during a session, because the current model does not provide good enough results --- such requests would fail with "no models available".

ronenkat (commented):
Session affinity is an important feature and should be made compatible with IPP which chooses models (which are fewer) and not vLLM nodes.

please also:
1. add a readme for the plugin.
2. document code at a function level.

pkg/framework/plugins/modelselector/filter/sessionaffinity/plugin.go
szedan-rh: you're right that this distinction matters.
 
  I was thinking about this from a server-level KV cache perspective, but as you point out, IPP operates on models/inference pools, not individual vLLM nodes. Hashing session IDs against model names doesn't
  achieve the cache locality I described in the PR — that requires pinning at the server level, which is the router's domain.

  Two questions to align on the right path forward:
  
  1. Is there a model-level session affinity use case worth solving here? For example, biasing repeat sessions toward the same model for consistency — not cache locality. If so, a scorer (soft preference) makes
  more sense than a filter (hard elimination), since it composes with other signals and degrades gracefully.
  2. Or should session affinity live entirely at the router layer? If the real value is KV cache reuse, that's a server-level concern and IPP isn't the right place for it. Happy to move this downstream if that's
  the better architecture.

  Let me know which direction makes sense and I'll rework accordingly.

ronenkat: The model-selector can choose a model from a set of candidate models.
For multi-turn conversations when there are multiple choices we should favor a previously selected model - so creating a bias toward hitting the same model (and not choosing another) make sense.
It should enable freedom on the first conversation turn, and allow potential KV-cache hits on follow-up turns.

ronenkat:
Please document requirements, for example if storing session id beyond the request context is not allowed, etc..

pkg/framework/plugins/modelselector/filter/sessionaffinity/plugin.go
szedan-rh: Thank you for clarifying that, I will change the implmentation.

szedan-rh: since IPP is stateless, the scorer needs to know which model was selected on previous turns. I'm thinking of echoing the selected model name back as a response header (e.g.
  X-Session-Model), and expecting the client to send it on the next request. That keeps us stateless with no server-side session storage. Does that approach work, or do you have a different mechanism in mind?

ronenkat: I am not sure that the custom header would be echoed back in the next turn --- there is no guarantee for it. 
This includes `X-Session-Id`.
If a session id exist, we should use it to track affinity.

szedan-rh: Thank you, I will change the implementation according to our discussion.

szedan-rh: Thanks for the feedback @ronenkat — here's a summary of the changes based on our discussion:
  
  Rewritten as a Scorer (not a Filter)

  The session affinity plugin is now a modelselector.Scorer under pkg/framework/plugins/modelselector/scorer/sessionaffinity/. It returns weighted scores instead of hard-filtering models, so it composes naturally with other scorers like inflight-requests-scorer via configurable weights.

  Key behaviors:
  - No session ID → all models score 0.5 (neutral), other scorers decide freely — first turn has full freedom
  - Session ID present → consistent hashing maps it to a preferred model (1.0), others get 0.0 — creates bias toward the same model for KV cache hits on follow-up turns
  - Preferred model removed → consistent hashing remaps only affected sessions (~1/N disruption), no errors

  Stateless, no client echo dependency

  Per your feedback that custom response headers won't be echoed back by clients — the plugin no longer reads or writes X-Session-Model, doesn't generate UUIDs, and doesn't implement ResponseProcessor. The session-to-model mapping is fully deterministic via consistent hashing: same session ID + same model set = same preferred model, with no stored state.

  The session ID is expected to be provided natively by the client (e.g. X-Session-Id header). If absent, the plugin has zero effect.

  Also added:
  - Plugin README with configuration and wiring docs
  - Function-level doc comments

Will update the PR in few

ronenkat: Let me suggest some updates:
1. no session id -> no opinion -> i would suggest a score of 0.0
2. session id present, get the last selection model (if still valid) using the session id.
   - if first turn, no model for session id --> no opinion
   - if session id provides a hint on last selected model, score the model with 1.0, all others with 0.0
3. You probably need to use the ResponseProcessor to track the mapping between session id and model.
Regarding configuration, enable session id lookup string, e.g., x-session-id, session_id. Provide a default set of string to serve as a session identifier

ronenkat: Please also echo back the session id in the Response when generated.
It would allow client who does echo back the content of the response header to take advantage of the session semantic.

szedan-rh: Thanks @ronenkat, I've reworked the implementation based on your suggestions:
 
  Scoring logic:
  - No session ID in request → score 0.0 (no opinion)
  - Session ID present, first turn (no prior model known) → score 0.0 (no opinion)
  - Session ID present, prior model known and still a candidate → score 1.0 for that model, 0.0 for others
  - Prior model no longer in candidates → score 0.0 (no opinion, graceful fallback)

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,210 @@
...
+		if val := request.Headers[key]; val != "" {
+			return val
+		}
+	}
+	return ""
+}
+
+// evictOldest removes roughly half the entries when the map hits capacity.
+// Called with s.mu held.
+func (s *SessionAffinityScorer) evictOldest() {
```
ronenkat: Please use LRU and TTL to manage session ids, instead of randomly evict half the list.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,210 @@
+/*
```
ronenkat, line 1: No metrics/observability - The plugin doesn't expose metrics for session hit rate, eviction frequency, or session map size.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,210 @@
...
+	modelName, ok := response.Body["model"].(string)
+	if !ok || modelName == "" {
+		return nil
+	}
+
+	s.mu.Lock()
+	if len(s.sessions) >= s.maxSessions {
+		s.evictOldest()
+	}
+	s.sessions[sessionID] = modelName
```
ronenkat: should be updated only if session is new or the model has changed

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,210 @@
...
+	s.mu.Unlock()
+
+	log.FromContext(ctx).V(logutil.VERBOSE).Info("recorded session affinity",
+		"sessionId", sessionID, "model", modelName)
+	return nil
```
ronenkat, line 256: please echo back the session-id in the response headers if not already there.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,210 @@
...
+	sessions map[string]string // session ID → model name
+}
+
+// TypedName returns the type and name tuple of this plugin instance.
+func (s *SessionAffinityScorer) TypedName() plugin.TypedName {
+	return s.typedName
+}
+
+// Score returns a score in [0,1] for each model based on session affinity.
+func (s *SessionAffinityScorer) Score(ctx context.Context, cycleState *plugin.CycleState, request *requesthandling.InferenceRequest, models []datalayer.Model) map[datalayer.Model]float64 {
```
ronenkat, line 175: Silent failures - Both `Score()` and `ProcessResponse()` silently skip errors, making debugging harder.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	return &SessionAffinityScorer{
+		typedName: plugin.TypedName{
+			Type: PluginType,
+			Name: name,
+		},
+		sessionIDKeys: config.SessionIDKeys,
+		maxSessions:   config.MaxSessions,
+		sessions:      make(map[string]string),
```
ronenkat: sessions should be captured in the datastore to be cross nodes

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin_test.go
```diff
@@ -0,0 +1,332 @@
...
+	models := make([]datalayer.Model, len(names))
+	for i, name := range names {
+		models[i] = datalayer.NewModel(name)
+	}
+	return models
+}
+
+// --- Factory tests ---
+
+func TestFactory_DefaultConfig(t *testing.T) {
```
ronenkat, line 54: please add a one-line explaining what is the goal of the test.

ronenkat (commented):
Thank you.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
szedan-rh: Agree with you, this is a significant scope increase. The current Datastore interface (datalayer.Datastore) is designed for model metadata, not arbitrary key-value storage. Adding session tracking to it would require
  either extending the interface or using model attributes in a non-standard way. I think this is worth discussing as a follow-up rather than blocking this PR. 
 what do you think?

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
szedan-rh: Done.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
szedan-rh: Done.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
szedan-rh: Done.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin_test.go
szedan-rh: Done.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
szedan-rh: 4 Prometheus metrics were added in plugin.go, registered via init() with the controller-runtime metrics registry, following the same pattern as pkg/metrics/metrics.go

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
ronenkat: sure. we can extend in a follow-up

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,273 @@
...
+	return s.typedName
+}
+
+// Score returns a score in [0,1] for each model based on session affinity.
+func (s *SessionAffinityScorer) Score(ctx context.Context, cycleState *plugin.CycleState, request *requesthandling.InferenceRequest, models []datalayer.Model) map[datalayer.Model]float64 {
+	logger := log.FromContext(ctx)
+	scores := make(map[datalayer.Model]float64, len(models))
+
+	sessionID, matchedKey := s.findSessionID(request)
+	if sessionID == "" {
```
ronenkat, line 180: please add an optimistic approach, where you add a session id, so if the client echo it back, we can track it.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
szedan-rh: https://github.com/llm-d/llm-d-inference-payload-processor/issues/187

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,273 @@
...
+func (s *SessionAffinityScorer) ProcessResponse(ctx context.Context, cycleState *plugin.CycleState, response *requesthandling.InferenceResponse) error {
+	logger := log.FromContext(ctx)
+
+	sessionID, err := plugin.ReadCycleStateKey[string](cycleState, cycleStateSessionIDKey)
+	if err != nil {
+		logger.V(logutil.VERBOSE).Info("no session ID in cycle state, skipping response processing")
+		return nil
+	}
+
+	modelName, ok := response.Body["model"].(string)
```
ronenkat: Make model a constant. E.g, 
`
requestModelField = "model"
`

pkg/framework/plugins/modelselector/scorer/sessionaffinity/sessioncache.go
```diff
@@ -0,0 +1,129 @@
...
+func (c *sessionCache) removeLocked(elem *list.Element) {
+	entry := elem.Value.(*sessionEntry)
+	delete(c.items, entry.sessionID)
+	c.order.Remove(elem)
+	if c.onEvict != nil {
+		c.onEvict()
+	}
+}
+
+func (c *sessionCache) evictLeastRecentLocked() {
```
ronenkat: We can use TTL here to remove multiple entries that passed their TTL period instead of just one.
please add a minEvictQuantity parameter that can be configured so remove a block of entries in order to reduce eviction events.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/sessioncache.go
```diff
@@ -0,0 +1,129 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	elem, ok := c.items[sessionID]
+	if !ok {
+		return "", false
+	}
+
+	entry := elem.Value.(*sessionEntry)
+	if c.nowFunc().Sub(entry.lastUsed) > c.ttl {
```
ronenkat: Get implies that we will use the information, so we should not remove it here.

ronenkat (commented):
Nice.
We should add a stress test for the session cache to validate that there is no contention on the cache when updating the session cache. Can be in a follow-up.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
szedan-rh: Done. When no session ID is present in the request, ProcessResponse now generates a UUID and echoes it back via the first configured header key (e.g. x-session-id). The mapping is stored in the cache  immediately. If the client echoes it on the next request, session affinity kicks in automatically. Added end-to-end test covering the full lifecycle: no ID → optimistic generation → client echo → affinity applied.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
szedan-rh: Done. Extracted responseModelField = "model" constant and replaced all usages.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/sessioncache.go
szedan-rh: Done. Eviction is now batch-based: when the cache hits capacity, it first sweeps all TTL-expired entries from the tail, then continues removing LRU entries until at least minEvictQuantity (configurable, default 5) total entries have been evicted. This amortizes eviction cost and reduces the frequency of eviction events. Added minEvictQuantity as a configurable parameter.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/sessioncache.go
szedan-rh: Done. Get() now returns ("", false) for expired entries without removing them — it's a pure read with no side effects. Expired entries are cleaned up during the batch eviction sweep in Put() when the cache reaches capacity.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/sessioncache.go
```diff
@@ -0,0 +1,158 @@
...
+// evictBatchLocked removes all TTL-expired entries from the tail, then
+// continues removing LRU entries until at least minEvictQuantity total
+// entries have been evicted.
+func (c *sessionCache) evictBatchLocked() {
+	now := c.nowFunc()
+	removed := 0
+
+	// Pass 1: sweep all TTL-expired entries from the tail
+	elem := c.order.Back()
+	for elem != nil {
```
ronenkat: please cap the loop with `10 x minEvictQuantity` to handle the case of large number of expired entries. In such case, we won't need to evict using LRU

pkg/framework/plugins/modelselector/scorer/sessionaffinity/sessioncache.go
```diff
@@ -0,0 +1,158 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	elem, ok := c.items[sessionID]
+	if !ok {
+		return "", false
+	}
+
+	entry := elem.Value.(*sessionEntry)
+	if c.nowFunc().Sub(entry.lastUsed) > c.ttl {
```
ronenkat: please remove the use of TTL in the get.
If we have the entry, lets use it. The use of TTL can be used to help remove entries.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/sessioncache.go
szedan-rh: Done. Eviction is now batch-based in sessioncache.go:evictBatchLocked() — when the cache hits capacity, it runs two passes:
  1. Sweeps all TTL-expired entries from the tail first
  2. Continues removing LRU entries until at least minEvictQuantity total entries have been evicted.

minEvictQuantity is configurable via the plugin parameters (default: 5). See SessionAffinityScorerConfig.MinEvictQuantity in plugin.go and sessionCache.minEvictQuantity in sessioncache.go

pkg/framework/plugins/modelselector/scorer/sessionaffinity/sessioncache.go
szedan-rh: Done. Get() in sessioncache.go now returns ("", false) for expired entries without removing them — pure read, no side effects. Expired entries are cleaned up during the batch eviction sweep (evictBatchLocked()) triggered by Put() when the cache reaches capacity.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/sessioncache.go
```diff
@@ -0,0 +1,158 @@
...
+		return "", false
+	}
```
ronenkat:
```suggestion
```

pkg/framework/plugins/modelselector/scorer/sessionaffinity/sessioncache.go
ronenkat:
```suggestion
	for elem != nil  && removed < c.minEvictQuantity * 10 {
```

ronenkat (commented):
Added suggestions inline

ronenkat (approved):
Thank you.

ronenkat:
@nirrozenbaum please take a look.
Follow up noted:
1. Tracking session affinity across IPP pods - #187 
1. Stress test on session tracking cache

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,282 @@
...
+func init() {
+	metrics.Registry.MustRegister(sessionCacheHits)
+	metrics.Registry.MustRegister(sessionCacheMisses)
+	metrics.Registry.MustRegister(sessionCacheSize)
+	metrics.Registry.MustRegister(sessionCacheEvictions)
+}
+
+// SessionAffinityScorerConfig defines the JSON configuration for the plugin.
+type SessionAffinityScorerConfig struct {
+	SessionIDKeys    []string `json:"sessionIdKeys"`
```
shmuelk: I understand the desire to support multiple session ID header values at once.

But I will point out that the EPP only supports one session ID header.

szedan-rh: Good point — noted. The multi-key lookup with defaults was requested by @ronenkat to accommodate different client conventions without requiring config changes. Since IPP and EPP serve different layers (model selection vs server routing), supporting multiple keys here adds flexibility at the IPP level. But agreed that operators should align the configured session ID key across both components for consistent end-to-end behavior. We can document this in the README as a deployment consideration.

szedan-rh: We  can also solve it by  add multi-key support to EPP (align with IPP), unless you think it's better to reduce IPP to a single key

shmuelk: I think its better to reduce the IPP to a single key

shmuelk:
Instead of writing your own LRU cache with a TTL expiration, why not use:

"github.com/hashicorp/golang-lru/v2/expirable"

szedan-rh:
Agree with you, no reason, wasn't aware about it.
i'm change the implmentation

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
szedan-rh: Thank you, handled.

shmuelk:
@szedan-rh Thanks for the changes

/lgtm
/approve

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,257 @@
...
+const defaultSessionIDKey = "x-session-id"
+
+var (
+	_ modelselector.Scorer              = &SessionAffinityScorer{}
+	_ requesthandling.ResponseProcessor = &SessionAffinityScorer{}
+)
+
+// --- Prometheus metrics ---
+
+var (
```
nirrozenbaum, line 63: nit: would be good if you can extract all metrics related vars/functions to `metrics.go` file under the same dir (align with llm-d-router structures)

nirrozenbaum: in a follow up

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,257 @@
...
+// When no session ID was present in the request, a new UUID is generated
+// optimistically and echoed back — if the client sends it on the next request,
+// session affinity will be established automatically.
+func (s *SessionAffinityScorer) ProcessResponse(ctx context.Context, cycleState *plugin.CycleState, response *requesthandling.InferenceResponse) error {
+	logger := log.FromContext(ctx)
+
+	modelName, ok := response.Body[responseModelField].(string)
+	if !ok || modelName == "" {
+		logger.V(logutil.VERBOSE).Info("no model in response body, skipping response processing")
+		return nil
```
nirrozenbaum, line 237: not all model servers return the model in the response.
in such a case we won't retain the session stickiness. 

we should maybe use some conventions here - the model selector always write to CycleState the selected model via `x-gateway-selected-model` key. then model selector can inject it during the request processing and we don't need to rely on structure of the response.

pkg/framework/plugins/modelselector/scorer/sessionaffinity/plugin.go
```diff
@@ -0,0 +1,257 @@
...
+	sessionID, err := plugin.ReadCycleStateKey[string](cycleState, cycleStateSessionIDKey)
+	if err != nil {
+		// No session ID in request — generate one optimistically
+		sessionID = uuid.New().String()
+		logger.V(logutil.VERBOSE).Info("generated optimistic session ID",
+			"sessionId", sessionID, "model", modelName)
+	}
+
+	s.cache.Add(sessionID, modelName)
+	sessionCacheSize.WithLabelValues().Set(float64(s.cache.Len()))
```
nirrozenbaum, line 249: if this is expensive operation consider making it async

nirrozenbaum (commented):
@szedan-rh thanks for the PR, overall it looks good. 
I left few minor comments that might be addressed in a follow up. 
also, I had one comment that was shared also in slack - 
if the session is returned in the initial request and then we expect the client to put it in followup requests..
then -
why don't we just say that the followup requests will put the selected model directly in the body (in model field)?
why do we keep sending `model: "auto"` with the selected model in a header?

if you agree with the above, maybe we can use a dedicated header and not the same one EPP uses. e.g., something like `X-Gateway-Selected-Model`. 
so when a client sends a request with `model: "auto"` , they should expect this header back and then inject it in the body in follow up requests, and the `X-Gateway-Session-ID` remains preserved for EPP, for selecting the endpoint.

PR #179, opened 2026-06-18
feat: add env var support to payload-processor Helm chart

yossiovadia:
## Summary

Add `env: []` field to the payload-processor Helm chart values for injecting custom environment variables into the container. Follows the same pattern used by [llm-d-router charts](https://github.com/llm-d/llm-d-router/blob/main/config/charts/routerlib/values.yaml#L37).

This enables downstream consumers to pass configuration like `GATEWAY_NAME` and `GATEWAY_NAMESPACE` to controllers registered via `WithCustomControllers`.

```yaml
payloadProcessor:
  env:
    - name: GATEWAY_NAME
      value: "my-gateway"
    - name: GATEWAY_NAMESPACE
      value: "default"
```

## Changes

- `values.yaml`: new `env: []` field with example
- `payload-processor.yaml`: passthrough `toYaml` in container spec

10 lines added, 2 files.

## References

- opendatahub-io/ai-gateway-payload-processing#310 — uses env vars for controller gateway config
- opendatahub-io/ai-gateway-payload-processing#334 — tracking issue


config/charts/payload-processor/templates/payload-processor.yaml
```diff
@@ -65,6 +65,9 @@ spec:
         - name: OTEL_TRACES_EXPORTER
           value: otlp
         {{- end}}
+        {{- if .Values.payloadProcessor.env }}
+        {{- toYaml .Values.payloadProcessor.env | nindent 8 }}
+        {{- end }}
```
nirrozenbaum: that's the wrong place. this should be under `env` section in line 37

config/charts/payload-processor/values.yaml
```diff
@@ -14,6 +14,13 @@ payloadProcessor:
     # Log verbosity
     v: 3
 
+  # Additional environment variables for the payload processor container.
+  # Example:
+  # env:
+  #   - name: GATEWAY_NAME
+  #     value: "my-gateway"
+  env: []
```
nirrozenbaum, line 22: can you add doc also in the helm README.md?

config/charts/payload-processor/templates/payload-processor.yaml
yossiovadia: Moved — now right after the NAMESPACE block inside the env: section. Verified with helm template.

config/charts/payload-processor/values.yaml
yossiovadia: Added to the configuration table in README.md.

PR #183, opened 2026-06-21
fix: Enable log and trace correlation

### LINKED ISSUE #161: [observability] Logs are not correlated with traces: trace_id/span_id never injected into the logger, aligned with llm-d-router#1630
/kind feature

## Description

IPP emits OpenTelemetry spans and structured zap (logr) logs, but the two are never correlated: no code path injects `trace_id`/`span_id` into the logger, and the zap setup has no automatic correlation hook. When debugging a request, there is no way to jump from a log line to the corresponding trace, or to find the logs belonging to a given trace.

This aligns IPP with llm-d/llm-d-router#1630, which tracks the same gap in the EPP and the disaggregation sidecar. EPP, IPP and the sidecar appear in the same traces (see #157), so log–trace correlation should be consistent across all three components.

## Current behavior

In the ext_proc entry point (`pkg/handlers/server.go`), `Process()` starts the `gateway.request` server span and only then retrieves the logger via `log.FromContext(ctx)`. Although the span context is already present in `ctx` at that point, the logger is never enriched with `trace_id`/`span_id` and is never stored back into the context, so every log line emitted during request processing (handlers, plugins, model selector) carries no trace correlation fields.

The global zap configuration (`pkg/common/observability/logging/logger.go`) is a plain JSON encoder with no OTel correlation hook, so there is no fallback either.

## Proposed change

In `Process()`, after the span is started, derive the request logger from the span context and store it back, e.g.:

    sc := span.SpanContext()
    if sc.IsValid() {
        logger = logger.WithValues(
            "trace_id", sc.TraceID().String(),
            "span_id", sc.SpanID().String(),
        )
        ctx = log.IntoContext(ctx, logger)
    }

All downstream code already retrieves the logger from the context, so this single injection point covers handlers, plugins and the model selector.

Whatever field names and mechanism llm-d-router#1630 settles on should be adopted here as-is, so that logs from EPP, IPP and the sidecar can be queried uniformly by `trace_id`.

## Why is this needed

- Without correlation, debugging a slow or failed request means manually matching timestamps between Jaeger and the log backend.
- The data layer makes IPP stateful (#85); diagnosing state-related issues will increasingly require connecting a specific request's trace to the log lines it produced.
- Router-side dashboards and runbooks being built around trace-correlated logs (llm-d-router#1630) will otherwise not work for the IPP hop of the same trace.

gyliu513:
**What type of PR is this?**

/kind bug

Fixes #161 

/cc @nirrozenbaum 

```release-note
Enabled log and trace correlation.
```

### DIFF
--- a/pkg/handlers/server.go
+++ b/pkg/handlers/server.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	extProcPb "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+	"github.com/go-logr/logr"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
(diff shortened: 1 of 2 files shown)


nirrozenbaum:
/lgtm
/approve

Thanks!

PR #186, opened 2026-06-22
feat: add active health checks for model endpoints

### LINKED ISSUE #182: Add active health checks for model endpoints with automatic withdrawal and recovery

Active, configurable health checks for model endpoints in the AI Gateway (IPP) that:

1. Probe model endpoints in the background via GET /v1/models (also supports TCP and HTTP/HTTPS GET)
2. Automatically exclude (withdraw) unhealthy models from the gateway's routing decisions after a configurable number of consecutive probe failures
3. Automatically re-include (recover) models once they start responding successfully again — no manual intervention required
4. Support configurable interval, timeout, unhealthy threshold, and healthy threshold per model
5. Assume external third-party models (e.g., OpenAI) are always healthy when no health endpoint is configured

**Why is this needed**:

Currently, if a model endpoint goes offline (server crash, OOM, network issue), the AI Gateway has no way to detect it and continues routing traffic to the dead endpoint — causing user-facing errors until someone manually intervenes.

- "The system must support active, configurable synthetic health checks (e.g., TCP port checks, HTTP/HTTPS GET requests) directed at the cluster's external load balancer IP."
- "If a specific AI cluster or model endpoint goes offline, the system must automatically withdraw that IP from the active pool within a configurable Time-To-Live (TTL) threshold."

This feature closes the gap by adding health check probes to the gateway layer (IPP) and impacting the gateway's model selection decision based on health status — ensuring unhealthy endpoints are never routed to, and recovered endpoints are automatically brought back into rotation.

abdallahsamabd:
## What type of PR is this?

/kind feature

## What this PR does / why we need it:

Adds active health checks for model endpoints in the AI Gateway (IPP) with automatic model server discovery from InferencePools. The health check system periodically probes internal model endpoints using `GET /v1/models` to verify both server liveness and model availability. Models that fail consecutive health checks are automatically withdrawn from routing, and are re-included once they recover.

## Key changes:

- **HealthCheckDataSource plugin:** Runs background probes per model at a configurable interval. Uses `ModelHTTPChecker` that validates the model name is present in the `/v1/models` response (not just a 2xx status), catching cases where the server is up but the model is unloaded (e.g. OOM). Dynamically starts/stops probes via a reconcile loop that reads config from the Datastore.
- **HealthFilter plugin:** Reads health status from the Datastore and excludes unhealthy models from the candidate list during model selection. Models with no health attribute (e.g. external models like OpenAI) pass through by default.
- **k8s-model-discovery-datasource plugin:** Automatically discovers model server pods by watching `InferencePool` CRDs — uses `spec.selector` to find matching pods and `spec.targetPorts` to determine endpoints. Probes each pod's `/v1/models` to discover which models are served, then writes `HealthCheckConfig` to the Datastore. No manual URL configuration required.
- **model-config-datasource extension:** Extended to support an optional `healthCheck` field per model in the `models.json` ConfigMap, allowing manual health check configuration as an alternative to auto-discovery.
- Configurable per-model: interval, timeout, healthy/unhealthy thresholds.
- External models are assumed always healthy and are not probed.

## Which issue(s) this PR fixes:

Fixes #182

## Release note (write NONE if no user-facing change):

Added active health check support for model endpoints with automatic discovery from InferencePools. The IPP discovers model server pods via InferencePool selectors and targetPorts, probes them via `GET /v1/models`, and automatically removes unhealthy models from routing decisions — re-including them upon recovery.

### DIFF
--- a/pkg/framework/plugins/datalayer/healthcheck/datasource.go
+++ b/pkg/framework/plugins/datalayer/healthcheck/datasource.go
@@ -1,0 +1,324 @@
+/*
+Copyright 2026 The llm-d Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package healthcheck
+
+import (
+	"context"
+	"encoding/json"
+	"sync"
+	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/datalayer"
+	dlsrc "github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/datalayer/datasource"
+	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/plugin"
+)
+
+// PluginType is the identifier used when registering this datasource.
+const PluginType = "health-check-datasource"
+
+// compile-time interface assertion
+var _ dlsrc.DataSource = &HealthCheckDataSource{}
+
+const defaultReconcileInterval = 5 * time.Second
+
+// DatasourcePluginConfig holds optional plugin-level defaults.
+type DatasourcePluginConfig struct {
+	ReconcileInterval string `json:"reconcileInterval,omitempty"`
+}
+
+// endpointState tracks consecutive successes/failures for one model.
+type endpointState struct {
+	consecutiveSuccesses int
+	consecutiveFailures  int
+	status               HealthStatus
+	cancel               context.CancelFunc
+}
+
+// HealthCheckDataSource discovers health check configuration from the Datastore
+// (written by model-config-datasource) and runs background probes accordingly.
+// It reconciles periodically to pick up additions/removals at runtime.
+type HealthCheckDataSource struct {
+	typedName         plugin.TypedName
+	ds                datalayer.Datastore
+	reconcileInterval time.Duration
+	states            map[string]*endpointState
+	mu                sync.RWMutex
+	stopCh            chan struct{}
+	doneCh            chan struct{}
+	wg                sync.WaitGroup
+}
+
+// DatasourceFactory creates a HealthCheckDataSource from plugin configuration.
+func DatasourceFactory(name string, rawCfg json.RawMessage, h plugin.Handle) (plugin.Plugin, error) {
+	reconcileInterval := defaultReconcileInterval
+
+	if rawCfg != nil && len(rawCfg) > 0 {
+		var cfg DatasourcePluginConfig
+		if err := json.Unmarshal(rawCfg, &cfg); err == nil && cfg.ReconcileInterval != "" {
+			if d, err := time.ParseDuration(cfg.ReconcileInterval); err == nil {
+				reconcileInterval = d
+			}
+		}
+	}
+
+	return &HealthCheckDataSource{
+		typedName:         plugin.TypedName{Type: PluginType, Name: name},
+		ds:                h.Datastore(),
+		reconcileInterval: reconcileInterval,
+		states:            make(map[string]*endpointState),
+		stopCh:            make(chan struct{}),
+		doneCh:            make(chan struct{}),
+	}, nil
+}
+
+func (d *HealthCheckDataSource) TypedName() plugin.TypedName { return d.typedName }
+
+// Start launches a reconcile loop that periodically scans the Datastore for
+// models with health check configuration and starts/stops probes dynamically.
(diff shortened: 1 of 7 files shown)


ronenkat:
@abdallahsamabd have you looked at the [model-config-datasource](https://github.com/llm-d/llm-d-inference-payload-processor/blob/main/pkg/framework/plugins/datalayer/modelconfigcollector/plugin.go#L51) plugin?
It provides a way to configure models from a config map, `CheckConfig` can be defined in the config-map and read via the config mechanism.
It would enable `CheckConfig` to be updated during runtime, which is not supported via the plugin parameters.
You should check if the config definition can fit into `ModelsConfig`, and the health checks/filter plugins will use the configuration which would be kept in the datastore to query the model endpoints.

abdallahsamabd:
Good point. I wasn't aware of the model-config-datasource pattern. You're right that plugin parameters are static and don't support runtime updates

pkg/framework/plugins/datalayer/healthcheck/datasource.go
```diff
@@ -0,0 +1,255 @@
...
+// CheckConfig defines a single endpoint health check.
+type CheckConfig struct {
+	Model              string   `json:"model"`
+	Type               string   `json:"type"` // "http" or "https"
+	URL                string   `json:"url"`
+	Interval           Duration `json:"interval"`
+	Timeout            Duration `json:"timeout"`
+	UnhealthyThreshold int      `json:"unhealthyThreshold"`
+	HealthyThreshold   int      `json:"healthyThreshold"`
+}
```
shmuelk: Do you expect the customer to know in advance the URLs of all of the model servers in a K8S cluster?

You need entries in this array for every single model server. This is not per pod, as the customer may be using vLLM's Data Parallel feature and run multiple model servers in the same pod.

You need some kind of automatic model server discovery process.

pkg/framework/plugins/datalayer/healthcheck/types.go
```diff
@@ -0,0 +1,39 @@
...
+limitations under the License.
+*/
+
+package healthcheck
+
+import "github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/datalayer"
+
+const (
+	// PluginType is the identifier used when registering this datasource.
+	PluginType = "health-check-datasource"
```
shmuelk: It is customary to place this constant in the source code file with the plugin.

abdallahsamabd: done, moved to datasource.go

pkg/framework/plugins/datalayer/healthcheck/datasource.go
abdallahsamabd: yes, in the current implementation (maybe needs to be changed...),

One pod (or Service) runs multiple models. Our health check handles this correctly because each model check validates its own name in the /v1/models response

config:
{
      "models": [
        {"name": "llama-8b",   "healthCheck": {"url": "http://vllm-svc:8000/v1/models"}},
        {"name": "mistral-7b", "healthCheck": {"url": "http://vllm-svc:8000/v1/models"}},
        {"name": "phi-3",      "healthCheck": {"url": "http://vllm-svc:8000/v1/models"}}
      ]
}
All 3 models point to the same URL. Each probe checks if its specific model is loaded.

for example:
GET http://vllm-svc:8000/v1/models
Response: {"data": [{"id":"llama-8b"}, {"id":"mistral-7b"}]}

llama-8b:   "llama-8b" found?   → healthy ✓
mistral-7b: "mistral-7b" found? → healthy ✓
phi-3:      "phi-3" found?      → NOT FOUND → unhealthy ✗


for automatic model server discovery, I'm planning to implement a new k8s-model-discovery-datasource plugin that watches InferenceModel + InferencePool CRDs, since those already exist as part of the GAIE deployment and contain the model-to-pool mapping. The plugin would:

1. Watch InferenceModel resources to discover model names and their poolRef
2. Resolve the pool's Service and port from InferencePool.spec.targetPortNumber
3. Auto-derive the health check URL: http://{pool-service}.{namespace}.svc:{port}/v1/models
4. Write health-check-config to the Datastore — consumed by the existing health-check-datasource unchanged

what do you think about this scenario?

shmuelk:
@abdallahsamabd said:

> One pod (or Service) runs multiple models. Our health check handles this correctly because each model check validates its own name in the /v1/models response

This is incorrect. A single vLLM instance can run only one model. It can run LoRA adapters that are based on that one model it is serving. See https://discuss.vllm.ai/t/run-multiple-models/1181

pkg/framework/plugins/datalayer/healthcheck/datasource.go
shmuelk: Several issues:

  1. The InferenceModel CRD has been deprecated and removed by the GAIE and the llm-d-router.
  2. The InferencePool does not have a service that points to the model servers. The service referenced is the service of the EPP and it communicates with Envoy over a "private" gRPC based protocol.
  3. Ditto for three above

ronenkat:
Is the goal of the health checks is to understand the `inference-pool` status? or the status of the serving pods?
I think the health checks should be for the llm-d router endpoint health status, not the actual vllm pods.

abdallahsamabd:
yes you are right,
I'll modify the approach to probe the llm-d router endpoint (pool-level) instead of individual vLLM pods. This way the health check answers "can this pool serve this model?" rather than "is this specific pod alive?"
btw this was the first approach, before trying to implement the automatic model discovery.....

PR #188, opened 2026-06-22
Queue TTFT Latency Scorer

### LINKED ISSUE #54: [Feature]: Latency-Based Model Routing
### Problem Statement

When multiple models are available for serving model requests, there is currently no intelligent routing mechanism to optimize request latency. This results in:

- No mechanism to determine optimal model to select
- Suboptimal request completion times when endpoints have varying loads
- Inefficient system utilization across available models and inference pools
- No consideration of queue depth, network latency, or endpoint performance characteristics

Administrators and users need a way to route inference requests to the endpoint that will provide the fastest response time, whether local or remote.

### Proposed Solution

Implement latency-based routing using the Filter→Scorer→Picker pipeline:

**Core Components:**

1. **Request-Response Tracking data source**: Background component computing moving averages for:
   - In-flight requests per endpoint
   - Time-to-First-Token (TTFT) and end-to-end latency
   - KV-cache hit potential via request prefix tracking
2. **Latency Scorers**: Non-blocking scorer that predicts latency based on:
   - Endpoint queue depth (i.e., concurrent requests)
   - KV-cache hit estimation (based on prefix affiliation)
   - Historical (e.g., sliding window style) TTFT and decode times

**Implement a set of scorers to compute various latency factors**:
- Serving endpoint queue tracking
- KV-cache hit estimation via prefix matching
- Request TTFT / completion time prediction score

**Requirements:**
- Pre-req: data layer to keep various topics "concurrency" and "prefix"
- Support in-cluster llm-d inference pools
- Can be extended to support external/remote serving
- No persistent state (in-memory only) - to be revisited at a later stage

### Alternatives Considered

- Simple round-robin routing (baseline for benchmarking)
- Static endpoint configuration without dynamic latency awareness
- External load balancer solutions (adds complexity and latency)

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

**Evaluation Plan:**
Benchmark scenarios comparing single pool, round-robin, and latency-based routing with varying pool configurations.

**Future Enhancements:**
- Add failure rate tracking

Mohammad-nassar10:
**What type of PR is this?**

Fixes #54.

```release-note
NONE
```


examples/plot_capacity.py
```diff
@@ -0,0 +1,141 @@
+"""
```
nirrozenbaum: is this file related to the PR?

examples/medianttft-values.yaml
```diff
@@ -0,0 +1,40 @@
...
+        lowLoadWindowAge: 1h         # window for two-level P10Low (long = stable hardware floor)
+    - type: model-config-datasource
+      parameters:
+        modelsPath: /config/models.json
+    profiles:
+    - name: default
+      plugins:
+        request:
+        - pluginRef: model-selector
+        - pluginRef: median-ttft-scorer
```
aradhalevy: This should be `queue-ttft-scorer` ?

pkg/framework/plugins/modelselector/scorer/queuettft/README.md
```diff
@@ -0,0 +1,86 @@
...
+```
+effectiveTTFT = P10Low + inflight x (P50 - P10Low) / inflightAtP50
+```
+Falls back to P10Low when P50 is not yet available or equals the floor.
+
+**Score:**
+```
+score = (maxTTFT - effectiveTTFT) / (maxTTFT - minTTFT)
+```
+Unobserved models score 1.0 (cold start) or 0.5 (idle alongside observed peers).
```
aradhalevy: Right now if a model is Unobserved and another is Observed (The problem also happens if the model wasn't used for the window we define, not only cold start), all the traffic will go to the Observed even if its really slow. I suggest giving the Unobserved a score of 1.0 so the system routes some requests to it, just so it will be observed.

pkg/framework/plugins/datalayer/ttftpercentile/plugin.go
```diff
@@ -0,0 +1,286 @@
...
+		s := e.state[model]
+		cutoff := now.Add(-e.windowAge)
+		for id, entry := range s.pending {
+			if entry.dispatchedAt.Before(cutoff) {
+				delete(s.pending, id)
+			}
+		}
+		m := s.TTFTPercentileMetrics
+		e.ds.GetOrCreateModel(model).GetAttributes().Put(AttributeKey, m)
+		p10eff := m.P10LowTTFT
```
ronenkat: is the following just for logging?

pkg/framework/plugins/datalayer/ttftpercentile/plugin.go
```diff
@@ -0,0 +1,286 @@
...
+				continue
+			}
+			model, _ := p.Request.Body["model"].(string)
+			if model == "" {
+				continue
+			}
+			s := e.getOrCreate(model)
+			inflight := s.Requests
+			s.Requests++
+			if reqID := p.Request.Headers["x-request-id"]; reqID != "" {
```
ronenkat: what is no "x-request-id" header is present?

pkg/framework/plugins/datalayer/ttftpercentile/plugin.go
```diff
@@ -0,0 +1,286 @@
...
+				s.pending[reqID] = pendingEntry{
```
ronenkat: why do you need the reqID? why not store in cycle-state?
if the issue is that it is not available, lets add it to event.

pkg/framework/plugins/datalayer/ttftpercentile/plugin.go
```diff
@@ -0,0 +1,286 @@
...
+				}
+				s.tracker.add(ttft, inflightAtDispatch, now)
+				debugLogger.Info("ttft-observation",
+					"model", model, "ttft_s", ttft, "inflightAtDispatch", inflightAtDispatch,
+				)
+			}
+			if reqID != "" {
+				delete(s.pending, reqID)
+			}
+			if now.Sub(s.intervalStart) >= e.intervalDuration {
```
ronenkat: Would P50 updates frequency be dependent on intervalDuration ?

ronenkat (commented):
The code is missing algorithm style documentation to help understand the intent of the code.

pkg/handlers/response.go
```diff
@@ -60,15 +60,20 @@ func (s *Server) HandleResponseHeaders(ctx context.Context, reqCtx *RequestConte
...
 	s.eventNotifier.Notify(datasource.Event{
 		Type: datasource.ResponseEventType,
 		Payload: datasource.ResponsePayload{
-			Request:  reqCtx.Request,
-			Response: reqCtx.Response,
-			Duration: reqCtx.ResponseCompleteTimestamp.Sub(reqCtx.RequestReceivedTimestamp),
-			TTFT:     reqCtx.ResponseFirstChunkTimestamp.Sub(reqCtx.RequestSentTimestamp),
+			Request:    reqCtx.Request,
+			Response:   reqCtx.Response,
+			CycleState: reqCtx.CycleState,
```
ronenkat, line 72: Extract adding CycleState to the Request/ResponsePayload payload to a separate PR

cmd/runner/runner.go
```diff
@@ -47,12 +47,16 @@ import (
 	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/datalayer/datasource"
 	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/plugin"
 	"github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/interface/requesthandling"
+	modelconfigcollector "github.com/llm-d/llm-d-inference-payload-processor/pkg/framework/plugins/datalayer/modelconfigcollector"
```
ronenkat, line 50: please rebase your branch

config/charts/payload-processor/templates/config.yaml
```diff
@@ -26,4 +26,11 @@ data:
     kind: PayloadProcessorConfig
     {{- .Values.payloadProcessor.customConfig | toYaml | nindent 4 }}
   {{- end }}
+  {{- if .Values.payloadProcessor.listModels }}
```
ronenkat: is this change related to the scorer?

pkg/framework/plugins/datalayer/ttftpercentile/plugin.go
Mohammad-nassar10: Good point. Changed to use cycle-state. Thanks!

pkg/framework/plugins/datalayer/ttftpercentile/plugin.go
Mohammad-nassar10: yes, it is updated every intervalDuration.

pkg/handlers/response.go
Mohammad-nassar10: Thanks!
Moved to PR #203.

pkg/framework/plugins/modelselector/scorer/ttftaware/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+
+func NewTTFTAwareScorer() *TTFTAwareScorer {
+	return &TTFTAwareScorer{
+		typedName:       plugin.TypedName{Type: PluginType, Name: PluginType},
+		explorationRate: defaultExplorationRate,
+	}
+}
+
+func (s *TTFTAwareScorer) TypedName() plugin.TypedName { return s.typedName }
+func (s *TTFTAwareScorer) WithName(name string) *TTFTAwareScorer {
```
nirrozenbaum, line 108: nit: missing space

```suggestion
func (s *TTFTAwareScorer) TypedName() plugin.TypedName { return s.typedName }

func (s *TTFTAwareScorer) WithName(name string) *TTFTAwareScorer {
```

pkg/framework/plugins/modelselector/scorer/ttftaware/plugin.go
```diff
@@ -0,0 +1,179 @@
...
+
+// metricsFor reads the TTFT percentile metrics an extractor published for the model.
+// A missing or malformed attribute yields the zero value, which reads as truly cold.
+func metricsFor(model datalayer.Model) ttftpercentile.TTFTPercentileMetrics {
+	if val, ok := model.GetAttributes().Get(ttftpercentile.AttributeKey); ok {
+		if m, ok := val.(ttftpercentile.TTFTPercentileMetrics); ok {
+			return m
+		}
+	}
+	return ttftpercentile.TTFTPercentileMetrics{}
```
nirrozenbaum: exactly for this use case we added `ReadAttributeKey` function:

```suggestion
	if val, err := datalayer.ReadAttributeKey[ttftpercentile.TTFTPercentileMetrics](model.GetAttributes(), ttftpercentile.AttributeKey); err == nil {
		return val
	}
	return ttftpercentile.TTFTPercentileMetrics{}
```

Mohammad-nassar10: fixed, thanks!

PR #199, opened 2026-06-29
feat: echo selected model name as response header

szedan-rh:
echo selected model name as response header  When model selection runs, the resolved model name is written to CycleState and echoed back to the client via X-Gateway-Model-Name response header, enabling clients to pin follow-up requests.


szedan-rh:
@nirrozenbaum - Could you please review?

pkg/framework/plugins/responsehandling/modelnametoheader/plugin.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (p *ModelNameToHeaderPlugin) ProcessResponse(ctx context.Context, cycleState *plugin.CycleState, response *requesthandling.InferenceResponse) error {
+	selectedModel, err := plugin.ReadCycleStateKey[string](cycleState, modelselector.SelectedModelCycleStateKey)
+	if err != nil {
+		log.FromContext(ctx).V(logutil.VERBOSE).Info("no selected model in CycleState, skipping")
+		return nil
+	}
+
+	response.SetHeader(bodyfieldtoheader.ModelHeader, selectedModel)
```
ronenkat: would reusing `X-Gateway-Model-Name` cause the proxy to re-route the request to the inference pool? or the proxy routing rule will ignore it?

pkg/framework/plugins/responsehandling/modelnametoheader/plugin.go
```diff
@@ -0,0 +1,65 @@
+/*
```
ronenkat, line 1: Can you document the plugin?
Include why it is needed, and what assumptions / requirement it makes for the user, e.g., reading the model name from the response and echoing it back. as this assumption is not based on any existing specification, and is newly defined here.

pkg/framework/plugins/responsehandling/modelnametoheader/plugin.go
nirrozenbaum: > would reusing X-Gateway-Model-Name cause the proxy to re-route the request to the inference pool?

no. this is done on the response (routing already completed).

cmd/runner/runner.go
nirrozenbaum, line 295: is this related to current PR?

pkg/framework/plugins/responsehandling/modelnametoheader/plugin.go
szedan-rh: Correct as Nir said — this runs on the response path, so routing is already complete. The proxy only uses X-Gateway-Model-Name on the request side for routing decisions.

pkg/framework/plugins/responsehandling/modelnametoheader/plugin.go
szedan-rh: Good point, will add documentation for the plugin.

cmd/runner/runner.go
szedan-rh: No, that's from #175 which already landed on main. I'll rebase to clean up the diff.

szedan-rh:
@nirrozenbaum / @ronenkat - Please take another look

PR #200, opened 2026-06-30
fix: remove chunk param from ResponseChunkProcessor, add mutated flag

noyitz:
## Summary
Follow-up to #169 addressing post-merge review comments:

- **Remove `chunk string` parameter** from `ProcessResponseChunk` interface — plugins read/write exclusively through `response.CurrentChunk` and `response.SetChunk()`. Eliminates ambiguity for implementors.
- **Pass `mutated` flag** through `buildStreamedChunkResponse` — reserved for future optimization to skip BodyMutation when chunk is unchanged, reducing envoy processing overhead.

## References
- https://github.com/llm-d/llm-d-inference-payload-processor/pull/169#discussion_r3476361302
- https://github.com/llm-d/llm-d-inference-payload-processor/pull/169#discussion_r3477574222
- https://github.com/llm-d/llm-d-inference-payload-processor/pull/169#issuecomment-4839711087

## Test plan
- [x] All existing tests pass
- [ ] Follow-up: add unit tests for chunk/body/no-plugin profile modes (tracked separately per Nir's request)

### DIFF
--- a/pkg/handlers/response.go
+++ b/pkg/handlers/response.go
@@ -137,7 +137,7 @@ func (s *Server) generateEmptyResponseBodyResponse(responseBodyBytes []byte) []*
 func (s *Server) HandleResponseChunk(ctx context.Context, reqCtx *RequestContext, chunkBytes []byte, endOfStream bool) ([]*eppb.ProcessingResponse, error) {
 	// Bodiless requests (e.g., GET /v1/models) may not have a profile set.
 	if reqCtx.Profile == nil || len(reqCtx.Profile.ResponseChunkProcessors) == 0 {
-		return s.buildStreamedChunkResponse(reqCtx, chunkBytes, endOfStream), nil
+		return s.buildStreamedChunkResponse(reqCtx, chunkBytes, endOfStream, false), nil
 	}
 
 	logger := log.FromContext(ctx).V(logutil.DEFAULT)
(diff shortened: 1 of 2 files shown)


pkg/handlers/response.go
```diff
@@ -150,12 +150,13 @@ func (s *Server) HandleResponseChunk(ctx context.Context, reqCtx *RequestContext
...
 
+	mutated := reqCtx.Response.ChunkMutated()
 	outBytes := chunkBytes
-	if reqCtx.Response.ChunkMutated() {
+	if mutated {
 		outBytes = []byte(reqCtx.Response.CurrentChunk)
 	}
 
-	return s.buildStreamedChunkResponse(reqCtx, outBytes, endOfStream), nil
+	return s.buildStreamedChunkResponse(reqCtx, outBytes, endOfStream, mutated), nil
```
nirrozenbaum: why do you pass mutated bool and in the function use wildcard (ignore it)?

noyitz: Good catch — this was addressed in the updated commit. The `mutated` bool was removed from `buildStreamedChunkResponse`'s signature since it wasn't being used inside the function. The current diff no longer passes it: `buildStreamedChunkResponse(reqCtx, outBytes, endOfStream)` with 3 params only. The chunk param was also removed from `ProcessResponseChunk` since it's already on `response.CurrentChunk`.

